PR #27, opened 2021-02-17
Configuration system change

TheTerrasque:
This creates a more modular and abstract, and more static typed configuration that doesn't rely on typing strings correctly everywhere. It can also take configuration from multiple sources, and should be easier to work with server-pushed configuration in a possible future feature.

To create a new section:
  1. Create a new file and class in Configurations/Sections folder
  2. Have it inherit BaseConfig with T of the new class created
  3. refer it in Configurations/Configuration.cs
  4. Add a line to load the section in Configfurations/ConfigurationExtra LoadIni function

To create a new option, either set it in the class, or refer to it where you need it and let VS create it for you. In most cases, VS can see which data type it should be and what access it should have.

To access a configuration entry use Configuration.Current.<Section>.<Property> - to check if the section is active use the bool Configuration.Current.<Section>.IsEnabled

I would like to get rid of step 4 for creating a new section, and would also like to have a shorter name path to refer to the configuration. But haven't found a good solution there yet. Maybe having a base class for handlers could help here.

Changes:

- All sections needs a "enabled=true" entry to be active. Previously that wasn't done for hotkeys
- Hotkeys for advanced building and editing mode is moved to their respective section. Previously they were all in hotkeys section
- Most settings have a default value set (all of 0.5 settings, some 0.6 settings) and can be omitted from the configuration if user doesn't want to change it.


nxPublic:
Thank you so much my man ❤️ 

I'll keep you updated on discord about questions that i might have 👍

PR #38, opened 2021-02-18
Simplify project setup

jokoho48:
**When merged this pull request will:**
- Simplify the Setup Procedure of the Project.
- Remove the need to add all References manually
- Adds Automatic Moving of the dll into the Plugins folder
- Allows the user to Start the game from the IDE (Sadly without Debugging ability)


nxPublic:
I will merge these changes appropriately after todays 0.8 release.

Greetings 😄

zedle:
This is a nice improvement, was considering having a 'Libs' folder in the project that you put the necessary deps into. Any reason you chose an env var over that approach @jokoho48 ?

jokoho48:
> 
> 
> This is a nice improvement, was considering having a 'Libs' folder in the project that you put the necessary steps into. Any reason you chose an env var over that approach @jokoho48 ?

I decided to use env var because with that you don't need duplicated files additionally, I added in an Automatic copy of the build files to the plugin folder so you can bind the Run/Build/Debug in the Editor directly to Valheim and after the build instantly (Aka less time wasted copying files manualy)

zedle:
👍  I would just change the readme to include a command line for setting the env var. 
eg `setx VALHEIM_INSTALL "C:\Program Files\Steam\steamapps\common\Valheim" /M`

jokoho48:
> 👍 I would just change the readme to include a command line for setting the env var.
> eg `setx VALHEIM_INSTALL "C:\Program Files\Steam\steamapps\common\Valheim" /M`

Good Point. I added it as an example below

TheTerrasque:
I like this PR, I was thinking of doing something to solve this problem, but with a different approach. I like his approach better.

On the step to copy the DLL, that should also include required dll's. ATM that's the inifileparser dll and probably by now the yamldotnet dll.

zedle:
👍

sirskunkalot:
I handled this in #99, but added more stuff on top. This could be closed

radmint (approved):

nxPublic (approved):

PR #79, opened 2021-02-22
AutoDeposit feature

Drakeny:
Addition of AutoDeposit feature for Smelters and Kilns
 - Smelters and Kilns will look for chests within specified range (AutoDepositRange) and search for available slots, if no slots are found within range it will spawn production on the ground (default behavior)

This one will need a bit more focused testing since there is possibility of duping or "deleting/not creating" player items, tested myself quite a bit but better safe than sorry.


zedle:
This is very cool but Smelter.cs needs some comments / whitespace imo

Drakeny:
> This is very cool but Smelter.cs needs some comments / whitespace imo

Thanks, check it out now, re-organized things a bit

zedle (approved):

nxPublic:
Great work! Thank you so much for contributing to V+ ❤️

PR #95, opened 2021-02-23
Larger inventory support

Zogniton:
* Can extend player inventory up to 20 x 8 from 4 x 8
  * ~~I could not add a scrollbar to the player inventory, 7 rows is the maximum you can fit while still seeing a chest inventory using 115% UI scale.~~
  * Added player inventory scrollbar thanks to Mixone
  * Up to 6 x 8 there is no scrollbar, 7 x 8 and up the scrollbar appears
* Can extend chest inventory (wood up to 10 x 8, iron up to 20 x 8)
  * Chests have a scrollbar if larger than 4 x 8
* Can change how inventory gets filled with materials (top to bottom or bottom to top)
* Tested;
  * Inventory Save/Load
  * Recover tombstone on death
  * 2 player Multiplayer (non dedicated)
    * same config works
    * joining with different config works but each player saw different size on chests
    * ~~**Potential issue**: server has smaller chests, player with larger chests puts something into the last slots, server does not save those items. Is there a way to not allow players to join if their config is different ?~~ I just found out about enforceConfiguration, we good


Zogniton:
image

Zogniton:
image

PR #99, opened 2021-02-24
Automatic publishing (local for Debug, zip for Release) + Debug compatibility for dnSpy

sirskunkalot:
**TL;DR**
You can stop copying shit around all the time

**What it does**
- I have added a PowerShell-script which gets executed via the project Build Event. Depending on the target it will copy the whole build into the Valheim game folder (Debug target) or assemble two zip files for Windows and Unix which are ready for release (Release target).
- BepInEx 5.4.6 is added as a binary to the project in order to package it for releases. Also added a reference in the project, so VS will copy the dll into the build target directory for easier distribution.
- Unix and Windows builds (including the Debug build for the local Valheim) both use the BepInEx-feature of a seperated folder with the unstripped dlls in it (atm this is just in the Unix release package). This leads to a easier publishing and an easier setup of this dev project (just add unstripped_corlib and publicized dlls to Valheim and you are sorted).
- Project references outside the distributable sphere (assembly, Unity.*, etc) and the publishing now rely on a csproj.user file for the path to the Valheim game folder (and probably more per-user-config in the future).
- Linux shell scripts for starting a game server with directly defined server arguments and the possibility to override any of them via script arguments (aka LinuxGameServerManager compatible).
- Compiled and added a unity-mono-dnspy dll used for debugging the game live via dnSpy
- A lot of general housekeeping

**What is missing**
- The new "dev package" one must extract into the local Valheim folder. Would upload that if this gets merged.
- Maybe more properties in the csproj.user for a finer control of the publishing process
- A smart solution for the valheim_plus.cfg for the local Debug build. One needs to be "ready-to-publish" (aka all default) and the game one needs obviously to be editied. It is "keep-track-yourself" at the moment (no config file is published locally in Debug)
- Maybe some more magic when having different valheim installs
- More sanity checks in the publish script
- ???
- Profit

Let me know what you think about it.


sirskunkalot:
By the way - as PowerShell can easily load and execute .NET assemblies, a much tighter integration of unittests or config validation could be integrated into that (e.g. generating the default valheim_plus.cfg). Seems more appropriate to me for a small project than completely integrating a new build system (ant or something like that).

sirskunkalot:
Totally forgot to mention: After the checkout you will have your csproj.user file deleted. This is on purpose, as this should be, well, per user and to add it to .gitignore I had to delete it from the repo first.

That being said, you will need a new one for this whole thing to function. Take this as a reference, edit to your local needs (Valheim path and Steam path) and save it as /ValheimPlus/ValheimPlus.csproj.user:

```xml
<?xml version="1.0" encoding="utf-8"?>
<Project ToolsVersion="Current" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <PropertyGroup>
    <!-- Needs to be your path to the base Valheim folder -->
    <ValheimDir>F:\Steam\steamapps\common\Valheim</ValheimDir>
  </PropertyGroup>
  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Debug|AnyCPU'">
    <!-- Runs Steam opening Valheim after building debug -->
    <StartAction>Program</StartAction>
    <StartProgram>C:\Program Files %28x86%29\Steam\steam.exe</StartProgram>
    <StartArguments>-applaunch 892970</StartArguments>
  </PropertyGroup>
</Project>
```

Restart Visual Studio so it can load the new file.

You will also need unstripped_corlib (from the V+ Unix package) in your Valheim folder. Current Windows release still ships with "valheim_Data" so it is possible you don't have that right now. I will append it to this comment (I hope that's okay?!). Just extract the files to your local Valheim installation:

[ucl.zip](https://github.com/valheimPlus/ValheimPlus/files/6038700/ucl.zip)

When both conditions are met (even if you have a clean, vanilla install of Valheim) you can go into VS, set Build to Debug and hit F5 for instant build, copy and starting the game.

Sorry for not stating that clearly before. README.MD in this branch has all the information for project setup but you should not be forced to find that information on your own ;)

SSimcox:
I find this very useful so I'm curious why you closed it. Personally it was large and hard to grok, but I tested it and found it useful.

radmint:
We really appreciated @sirskunkalot's work here but ultimately we thought that this feature would be better suited as a standalone tool since it could used by other mod devs and not just by us. It's also a large feature for us to maintain and update as part of the integral V+ mod.

sirskunkalot:
@SSimcox Basically what @radmint said. I cannot completely understand all of their points but then again I do not have insights into their internal processes. What I fully understand and totally agree with: The main feature is not bound to V+ but suitable for all Valheim modders needs I will release the core features of this standalone.

Plus: This PR was way to overloaded with other housekeeping commits, which should have been not in there (my bad).

SSimcox:
@sirskunkalot Sounds good! I look forward to seeing the standalone!

Issue #90, opened 2021-02-23
[Feature Request] Sleeping with multiple players


Xiaminou:
I would like to have something similar to Minecraft's Morpheus mod for server that have multiple players playing at all time.
Basically an option to chose the % of your players that need to be sleeping in order to skip the night.

GaelicGamer:
Hmm, this shouldn't be too hard, let me take a quick look at this.

GaelicGamer:
Yup got this working locally, just need to get some folk onto my server to test it with different numbers

nxPublic:
No idea what happened to the PR but I'll add it to the to do list I am working on right now non the less.

PR #103, opened 2021-02-24
Updating Unarmed Config and Adding Sleep Config for Number of Players

GaelicGamer:
Should address [Feature Request] Sleeping with multiple players #90
Also, split Unarmed back out of the player section and renamed it to better explain what it does, since it's not the base that it modifies more than the value that skill will help you scale toward.


ValheimPlus/Configurations/Configuration.cs
```diff
@@ -31,5 +31,7 @@ public class Configuration
         public CameraConfiguration Camera { get; set; }
         public GameConfiguration Game { get; set; }
         public VagonConfiguration Wagon { get; set; }
+        public SleepConfiguration Sleep { get; set; }
+        public UnarmedConfiguration Unarmed { get; set; }
```
radmint, line 35: I think for now we should keep unarmed configuration where it is until we have a large discussion about our convention on organizing settings

ValheimPlus/ObjectDB.cs
```diff
@@ -0,0 +1,64 @@
+﻿using HarmonyLib;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using UnityEngine;
+
+namespace ValheimPlus
```
radmint, line 8: Can we remove this file?

valheim_plus.cfg
```diff
@@ -232,6 +238,22 @@ dataRate=60
 ; The interval in seconds that the game auto saves at (client only)
 autoSaveInterval=1200
 
+[Sleep]
+enabled=false
+; Change false to true to enable this section
```
radmint, line 243: For convention, let's move the comments above the property declarations

ValheimPlus/Game.cs
```diff
@@ -60,5 +60,48 @@ private static Boolean Prefix(ref Game __instance, ref Vector3 pos, ref int __re
             return true;
         }
     }
+    
+    [HarmonyPatch(typeof(Game), "EverybodyIsTryingToSleep")]
+    public static class Game_EverybodyIsTryingToSleep
+    {
+        private static bool Postfix(bool _result, ref ZDO __instance)
```
radmint, line 67: If we make this a `Postfix` patch, this will double the amount of calls to get character ZDOs since we run the original and then run this and override that. We should look at moving to use `Prefix` and skip the original method since we have all of the original method code here anyways

valheim_plus.cfg
```diff
@@ -232,6 +238,22 @@ dataRate=60
...
+
+byPlayers= false
```
radmint, line 245: I think we should only use `byPercentage` for sleep configuration at least for now. Having these two will create some confusion and I think the percentage setting would be more valuable. What do you think?

ValheimPlus/Game.cs
```diff
@@ -60,5 +60,48 @@ private static Boolean Prefix(ref Game __instance, ref Vector3 pos, ref int __re
...
+        {
+            if (Configuration.Current.Sleep.IsEnabled)
+            {
+                if (Configuration.Current.Server.IsEnabled)
```
radmint, line 71: Why are we checking that Server is enabled here?

radmint (changes requested):
Thanks @GaelicGamer I've left feedback, please update your PR summary to include which test cases you were able to check. Have you tested this in conjunction with max player increase configuration?

ValheimPlus/Configurations/Configuration.cs
GaelicGamer: I disagree, placing it under player with an incorrect description doesn't seem correct to me.  It doesn't modify the base unarmed damage and was moved without that same discussion, I am moving it back to where I originally implemented it.

ValheimPlus/ObjectDB.cs
GaelicGamer: Yeah, it's mostly testing injection of new items, which I now have working locally so will be part of a later PR

valheim_plus.cfg
GaelicGamer: We should declare that somewhere since they were below them in previous iterations of the file.

ValheimPlus/Game.cs
GaelicGamer: Good point, I'll swap it around.

valheim_plus.cfg
GaelicGamer: I think the player is more valuable and only added the percentage based on discussion, and it is always more confusing for players to understand what 33% of 11 players is, but 3 players are always 3 players.

ValheimPlus/Configurations/Configuration.cs
Xiaminou: I agree with @GaelicGamer, I was going to create an issue to request that it be moved somewhere else for people like me who want to modify the Player section but don't want to modify Unarmed Damage.

ValheimPlus/Game.cs
GaelicGamer: Good point, removing the check.

valheim_plus.cfg
GaelicGamer: Regardless, once I'm able to it's an easy change, I can move them.

valheim_plus.cfg
GaelicGamer: That and they are both there for that exact purpose, you don't have to turn both on, you can only use one of them if you care to.

GaelicGamer:
> Thanks @GaelicGamer I've left feedback, please update your PR summary to include which test cases you were able to check. Have you tested this in conjunction with max player increase configuration?

Yeah, we did testing all the way up to 18 players on my server instance, both configurations of the sleep portion work as intended.

GaelicGamer:
Closed due to a rejection of the conflict of interest request I submitted within my company.

ValheimPlus/Humanoid.cs
```diff
@@ -8,13 +8,13 @@ public static class ModifyCurrentWeapon
     {
         private static ItemDrop.ItemData Postfix(ItemDrop.ItemData __weapon, ref Character __instance)
         {
-            if (Configuration.Current.Player.IsEnabled)
+            if (Configuration.Current.Unarmed.IsEnabled)
```
Mixone-FinallyHere, line 11: Use Player

ValheimPlus/Humanoid.cs
```diff
@@ -8,13 +8,13 @@ public static class ModifyCurrentWeapon
...
             {
                 if (__weapon != null)
                 {
                     if (__weapon.m_shared.m_name == "Unarmed")
                     {
-                        __weapon.m_shared.m_damages.m_blunt = __instance.GetSkillFactor(Skills.SkillType.Unarmed) * Configuration.Current.Player.baseUnarmedDamage;
+                        __weapon.m_shared.m_damages.m_blunt = __instance.GetSkillFactor(Skills.SkillType.Unarmed) * Configuration.Current.Unarmed.scaleDamage;
```
Mixone-FinallyHere, line 17: Use Player

ValheimPlus/ObjectDB.cs
```diff
@@ -0,0 +1,64 @@
+﻿using HarmonyLib;
```
Mixone-FinallyHere, line 1: Don't add add test code to PR

ValheimPlus/ValheimPlus.csproj
```diff
@@ -347,7 +347,9 @@
     <Compile Include="Configurations\Sections\PlayerConfiguration.cs" />
     <Compile Include="Configurations\Sections\ServerConfiguration.cs" />
     <Compile Include="Configurations\Sections\FireplaceConfiguration.cs" />
+    <Compile Include="Configurations\Sections\SleepConfiguration.cs" />
     <Compile Include="Configurations\Sections\TimeConfiguration.cs" />
+    <Compile Include="Configurations\Sections\UnarmedConfiguration.cs" />
```
Mixone-FinallyHere, line 352: Use Player

ValheimPlus/ValheimPlus.csproj
```diff
@@ -359,16 +361,17 @@
...
     <Compile Include="Experience.cs" />
     <Compile Include="Fireplace.cs" />
+    <Compile Include="Game.cs" />
     <Compile Include="Helper.cs" />
     <Compile Include="Humanoid.cs" />
     <Compile Include="Initialize.cs" />
     <Compile Include="Fermenter.cs" />
     <Compile Include="Inventory.cs" />
     <Compile Include="Items.cs" />
+    <Compile Include="ObjectDB.cs" />
```
Mixone-FinallyHere, line 371: Don't add add test code to PR

valheim_plus.cfg
```diff
@@ -206,8 +207,13 @@ baseAutoPickUpRange=2
 ; Disable all types of camera shake
 disableCameraShake=false
 
-; The base unarmed damage multiplied by your skill level
-baseUnarmedDamage=10
+[Unarmed]
```
Mixone-FinallyHere, line 210: Use Player

valheim_plus.cfg
```diff
@@ -206,8 +207,13 @@ baseAutoPickUpRange=2
...
+
+enabled=false
```
Mixone-FinallyHere, line 212: Unarmed is NOT a separate thing, it does not need to be enabled if Player not enabled

valheim_plus.cfg
Mixone-FinallyHere, line 245: Use percentage always, can introduce unnecessary issues, and if they don't know how a simple percent works, they should not be hosting servers. Period.

valheim_plus.cfg
```diff
@@ -232,6 +238,22 @@ dataRate=60
...
 autoSaveInterval=1200
 
+[Sleep]
+enabled=false
+; Change false to true to enable this section
+
+byPlayers= false
+; Change to true to sleep when a number of players are sleeping
+
+numberOfPlayersToSleep = 1
```
Mixone-FinallyHere, line 248: Use percentage always, can introduce unnecessary issues, and if they don't know how a simple percent works, they should not be hosting servers. Period.

valheim_plus.cfg
```diff
@@ -232,6 +238,22 @@ dataRate=60
...
+; The number of players required to initiate sleep
+
+byPercentage = false;
```
Mixone-FinallyHere, line 251: We only need one enabled, not two.

valheim_plus.cfg
```diff
@@ -232,6 +238,22 @@ dataRate=60
...
+; Change to true to sleep when a percentage of players are sleeping
+
+percentageOfPlayersToSleep = 50;
```
Mixone-FinallyHere, line 254: This is good

ValheimPlus/Configurations/Sections/UnarmedConfiguration.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
```
Mixone-FinallyHere, line 1: Use Player

ValheimPlus/Configurations/Sections/SleepConfiguration.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Text;
+
+namespace ValheimPlus.Configurations.Sections
+{
+    public class SleepConfiguration : ServerSyncConfig<SleepConfiguration>
+    {
+        public bool byPlayers { get; internal set; } = false;
+        public int numberOfPlayersToSleep { get; internal set; } = 1;
+        public bool byPercentage { get; internal set; } = false;
+        public double percentageOfPlayersToSleep { get; internal set; } = 50;
```
Mixone-FinallyHere, line 13: Only this is needed, and it does not need to be a double

ValheimPlus/Configurations/Sections/PlayerConfiguration.cs
```diff
@@ -6,7 +6,5 @@ public class PlayerConfiguration : ServerSyncConfig<PlayerConfiguration>
         public float baseMegingjordBuff { get; internal set; } = 150;
         public float baseAutoPickUpRange { get; internal set; } = 2;
         public bool disableCameraShake { get; internal set; } = false;
```
Mixone-FinallyHere, line 8: Unarmed damage is a property of Player, not its own config, use Player

ValheimPlus/Configurations/Configuration.cs
Mixone-FinallyHere: Unarmed damage is a property of Player

If you don't want to modify Unarmed damage then simply do not modify it

The same way that you can set hotkeys and values for other settings, unarmed damage is but a property of the Player

Furthermore, its code is used in the Humanoid, if ever we are to add any other modifier to Human, do we then create a setting for each? Jump distance ? Run speed ?

At most, a separate config for PlayerModifiers

But a whole setting class only for Unarmed makes no sense

Mixone-FinallyHere (changes requested):
The sleeping code you suggest is rather inefficient and as oos stated, it is simply redoing calculations

I have put together a snippet for the sleep check that should be clearer and more in line with the structure of this repository and project. Please implement it. 

```cs
[HarmonyPatch(typeof(Game), "EverybodyIsTryingToSleep")]
public static class Game_EverybodyIsTryingToSleep
{
    private static bool Prefix(bool _result, ref ZDO __instance)
    {
        if (Configuration.Current.Sleep.IsEnabled)
        {
            List<ZDO> allCharacterZdos = ZNet.instance.GetAllCharacterZDOS();                    
            int sleepingCount = allCharacterZdos.Count(x => x.GetBool("inBed"));
            if (sleepingCount == 0)
                return false;
            int percentSleeping = sleepingCount / allCharacterZdos.Count * 100;
            if (percentSleeping >= Configuration.Current.Sleep.percentageOfPlayersToSleep)
            {
                _result = true;
                return false;
            }                    
            _result = false;
            return false;
        }
    return true;                
    }
}
```

As it stands the code in the sleep check will be rather inefficient and has a couple of coding choices the team doesn't really understand, such as using a `double` to store the result of an incremental count and a percentage that will be entered not as `double` but rather as `int`.

And for the future, if a team member makes a code structure edit, such as changing the location of a class, please do not simply change it back to how you like it in a PR. If a decision we take is not clear we have no issue detailing why we chose to take a decision or another. Simply reverting a change to your preference without previously raising a point is not the way to go about it

ValheimPlus/Configurations/Configuration.cs
Xiaminou: > 
> 
> Unarmed damage is a property of Player
> 
> If you don't want to modify Unarmed damage then simply do not modify it
> 
> The same way that you can set hotkeys and values for other settings, unarmed damage is but a property of the Player
> 
> Furthermore, its code is used in the Humanoid, if ever we are to add any other modifier to Human, do we then create a setting for each? Jump distance ? Run speed ?
> 
> At most, a separate config for PlayerModifiers
> 
> But a whole setting class only for Unarmed makes no sense

Correct me if I'm wrong but your implementation actually modifies the way Unarmed Damage works, doesn't it?
At the very least the default value shipped with Valheim Plus' config is not the same as the base game's, I noticed a considerable difference when you guys implemented it. Which is to say there is no way right now to "not use it".

PR #109, opened 2021-02-24
Server sync checksum fix

TheTerrasque:
Fix problem with float having different string formatting on different systems / locales


zedle (approved):

nxPublic:
Not relevant anymore, system is removed.

❤️

PR #121, opened 2021-02-25
Sconce and Brazier fuel fix

vincentlau36:
Reason:
Sconce was spelled incorrectly and the brazier hover text was not "brazier" but "Fire", preventing the fuel set on update from working on these two items. 

Actions:
Corrected the spelling from "Scounce" to "Sconce".
Using instance name for the Hanging Brazier instead of hover text.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ We supply 4 different versions of V+ with every release since version 0.8. You c
 
 ## **[Server] Windows**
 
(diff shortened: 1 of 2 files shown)


radmint:
This issue has been fixed in the latest release already but thank you for the PR and feel free to make a PR for any future fixes you see necessary!

https://github.com/valheimPlus/ValheimPlus/blob/master/ValheimPlus/Fireplace.cs

vincentlau36:
Oops, didn't see it yet, Thank you!

PR #126, opened 2021-02-25
Larger inventory support

Zogniton:
* Can extend player inventory up to 20 x 8 from 4 x 8
  * Added player inventory scrollbar thanks to Mixone
  * Up to 6 x 8 there is no scrollbar, 7 x 8 and up the scrollbar appears
* Can extend chest inventory (wood up to 10 x 8, iron up to 20 x 8)
  * Chests have a scrollbar if larger than 4 x 8
* Can change how inventory gets filled with materials (top to bottom or bottom to top)
* Tested;
  * Inventory Save/Load
  * Recover tombstone on death
  * 2 player Multiplayer (non dedicated)


radmint:
Thanks Zognitron! This looks good and I'm merging it in so that we can test for the next release! 🎉

zedle (approved):

ValheimPlus/Inventory.cs
```diff
@@ -1,19 +1,132 @@
...
+                int woodChestRows = Math.Min(10, Math.Max(2, Configuration.Current.Inventory.woodChestRows));
+
+                // Iron chest number of columns, clamp value in range 6-8
+                int ironChestColumns = Math.Min(8, Math.Max(6, Configuration.Current.Inventory.ironChestColumns));
+
+                // Iron chest number of rows, clamp value in range 3-20
+                int ironChestRows = Math.Min(20, Math.Max(3, Configuration.Current.Inventory.ironChestRows));
+
+                // Wood chest
+                if (h == 2 && w == 5)
```
radmint, line 64: @Zogniton can we check on name equality instead of h & w? That way we can separate out iron chest from boat and cart inventory. [Check out how we do this in Fireplace.cs for convention.](https://github.com/valheimPlus/ValheimPlus/blob/84341c51fae2e9fd3e24428ef6ca561544d0b1f0/ValheimPlus/Fireplace.cs#L50)

ValheimPlus/Inventory.cs
```diff
@@ -1,19 +1,132 @@
...
+    [HarmonyPatch(typeof(Inventory), MethodType.Constructor)]
+    [HarmonyPatch(new Type[] { typeof(string), typeof(Sprite), typeof(int), typeof(int) })]
+    public static class changeInventorySlotCount
+    {
+        static void Prefix(ref string name, ref Sprite bkg, ref int w, ref int h)
+        {
+            if (Configuration.Current.Inventory.IsEnabled)
+            {
+                // Player inventory number of rows, clamp value in range 4-20
+                int playerInventoryRows = Math.Min(20, Math.Max(4, Configuration.Current.Inventory.playerInventoryRows));
```
radmint, line 49: let's move these clamps to consts on the class

README.md
```diff
@@ -296,6 +296,36 @@ noWeatherDamage=false
 ; The maximum range that you can place build objects at
 maximumPlacementDistance=5
 
+[Inventory]
```
with-heart, line 299: Worth mentioning this functionality further up in the `README.md` with the rest of the features!

ValheimPlus/Inventory.cs
```diff
@@ -1,19 +1,132 @@
...
+                return true;
+            }
+        }
+    }
+
+    [HarmonyPatch(typeof(Inventory), MethodType.Constructor)]
+    [HarmonyPatch(new Type[] { typeof(string), typeof(Sprite), typeof(int), typeof(int) })]
+    public static class changeInventorySlotCount
+    {
+        static void Prefix(ref string name, ref Sprite bkg, ref int w, ref int h)
```
radmint, line 44: We can follow convention here for any `ref` the parameter name should be prefixed with `__`. Ex: `ref string name` => `ref string __name`

valheim_plus.cfg
```diff
@@ -54,6 +54,36 @@ noWeatherDamage=false
 ; The maximum range that you can place build objects at
 maximumPlacementDistance=5
 
+[Inventory]
+; Change false to true to enable this section.
```
radmint, line 58: Thank you for the descriptive comments in the cfg file here 👍🏼

radmint (changes requested):
@Zogniton some small change requests. Post on Discord if you have any questions!

ValheimPlus/Inventory.cs
Zogniton: I cannot do that as i get a [C0133](https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs0133) error
I could not have them at all and just assign them to h and w. ie: h = Math.....

Zogniton: > let's move these clamps to consts on the class

I cannot do that as i get a [C0133](https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs0133) error
I could not have them at all and just assign them to h and w. ie: h = Math.....

ValheimPlus/Inventory.cs
Zogniton: I tried this initially and it could probably be done, there were 2 things i did not like about it:
1. Chest names seem to be inconsistent, sometimes you get "Chest" and "Inventory", other times you get $piece_chestwood
2. I felt that adding 4 more parameters in the config file is a bit much considering vanilla iron chests, cart and boats are the same size

I can still try to make this work if needed

ValheimPlus/Inventory.cs
Zogniton: > We can follow convention here for any `ref` the parameter name should be prefixed with `__`. Ex: `ref string name` => `ref string __name`

As far as i know this is not possible. When patching a class constructor the parameters must be exact.
Here is the error:
[Error  :  HarmonyX] Failed to patch System.Void Inventory::.ctor(System.String name, UnityEngine.Sprite bkg, System.Int32 w, System.Int32 h): System.Exception: Parameter __name does not contain a valid index

README.md
Zogniton: done, thanks

ValheimPlus/Inventory.cs
radmint: oh sorry I mean the '20', '4', values set to consts like:
```
private const int playerInventoryMaxRows = 20;
private const int playerInventoryMinRows = 4;
```

ValheimPlus/Inventory.cs
radmint: Hmm I'll look more into this

ValheimPlus/Inventory.cs
radmint: Oh I might be thinking of something else, sorry about that

ValheimPlus/Inventory.cs
Zogniton: ah, makes sense, done

radmint:
@Zogniton just a couple of merge conflicts, not sure where you and @Mixone-FinallyHere are on figuring out how y'all are getting this in. Feel free to ignore resolving these if y'all have decided to merge it in another way! Thanks again!

Zogniton:
Code has been merged to dev branch, closing

PR #127, opened 2021-02-25
Show all portals on map with their respective names

Algorithman:
Show all portals on map with their respective names (connected portals prefixed with a *)
Sync on login or rename (with 5sec delay to account for ZDOS transfer)


Algorithman:
Probably needs a configuration entry too, didn't look into config yet though. This was only planned for our server, but i think it's quite useful overall.

Algorithman:
Integrated into #151

PR #148, opened 2021-02-26
Change or disable skill lower factor

vincentlau36:
Added configuration (server synced) for Skills.
Added ability to change the skill lower factor on player death.


vincentlau36:
For this feature I wasn't sure if you are following the same class name convention. Their code uses "Skills" instead of the singular "Skill" so I named my configuration and files "Skills" as well. Let me know if that should change.

vincentlau36:
Made a mistake with this branch. I will correct and draft up another PR. Sorry about that

PR #151, opened 2021-02-26
Feature: iniseparation and revamp

Algorithman:
Separating the config into different ini files serves multiple purposes.

1. Find stuff easily
2. Add non-existant values automatically on save (coming from older versions for example)

Also, the separation lets VH+ decide what to save where.
For example: 
StaminaUsage values always have to be on the server and have to be synced to the clients.
But client-only values don't need to be in the server config at all.
This approach lets players make their own customization (for example enabling advanced building mode or not) and they won't be tempted to change any server values (which would have no effect anyway, since they are synced from the server).

Another point is, changing server values ingame (via admin console commands for example) and automatically syncing to the clients.
In the previous approach everything was transferred, not distinguishing between the two different categories.

And: No more delivery of a default config file. People don't need to backup their old config and add new values by hand.
It is automatically created client and server side.
If there's a new config section, no problem, it's just written with default values on start.

Each configuration value can have it's own comment and introduction version. -> easy to spot new values, just look for the latest version number.

Let me know what you think.

### DIFF
--- a/ValheimPlus/AdvancedEditingMode.cs
+++ b/ValheimPlus/AdvancedEditingMode.cs
@@ -34,11 +34,12 @@ namespace ValheimPlus
                 }
                 else
                 {
+                    /*  Disabled for now, the max zoom default values are not correct
                     // Set default values first
                     __instance.m_maxDistance = Configuration.Current.Camera.GetDefault<float>(nameof(Configuration.Current.Camera.cameraMaximumZoomDistance));
                     __instance.m_maxDistanceBoat=Configuration.Current.Camera.GetDefault<float>(nameof(Configuration.Current.Camera.cameraBoatMaximumZoomDistance));
                     __instance.m_fov = Configuration.Current.Camera.GetDefault<float>(nameof(Configuration.Current.Camera.cameraFOV));
-
+                    */
                     if(Configuration.Current.Camera.IsEnabled)
                     {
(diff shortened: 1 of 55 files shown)


Algorithman:
More to come, need to find a proper hook for game shutdown to save configs.

nxPublic:
Have you made any proper tests in practice with people on a server with these changes?

Algorithman:
me + my trusted VM for now, will try with a few friends this evening.

sirskunkalot:
I am up for testing, too. sounds like a very good improvement (no broetli, though ;)). I also have a linux vm on the nets. So we could test how server switching and syncing config again behave

Algorithman:
It's already a ZPackage (i assume the Z stands for zip, at least i hope so ^^)

Algorithman:
Todo:

- [x] Multiplayer check configuration value transfer on change
- [ ] Save on shutdown check
- [ ] Check configuration consumers if they need event system
- [ ] Implement event system
- [ ] Sort configuration, some values should be server side and vice versa

sirskunkalot:
Tested distribution, access restrictions, serialization - local and remote. No problems encountered.

Algorithman:
Please review

Algorithman:
Bunch of self-inflated egos selfcongratulating themselves all the time (in for the money i presume).
Code quality is amateur at best.
I don't want to contribute to this project.

PR #152, opened 2021-02-26
pvp allways on

dariuskis:
Hello there is it possible to do like allways people shown on map, and cant remove it , so is it impossible to do with pvp status, what everyone would be with pvp and wount be able to turn it off?


radmint:
hi @dariuskis please use the "Issue" feature in the future for "Feature Requests" this feature is on our roadmap.

PR #153, opened 2021-02-26
Improve code of Inventory

Mixone-FinallyHere:
Change to Prefix, if config is enabled and noTeleport set, we can just ignore the function so return false


radmint (approved):

Mixone-FinallyHere:
Have implemented in dev side

PR #159, opened 2021-02-26
Change or disable skill loss factor

vincentlau36:
Added configuration (server synced) for Skills.
Added ability to change the skill lower factor on player death.


Xiaminou:
The default skill loss is 5% not 25%.
Source: https://valheim.fandom.com/wiki/Skills#Skill_Decrease

radmint:
Thank you so much for this contribution but this has already been implemented on our release branch and will be in the next release.

PR #160, opened 2021-02-26
Fix to weapons and tools stamina changes affecting all stamina drain sources

AccretionCD:
Very small addition as per the title


nxPublic:
The function it self is not working as intended at the moment.

It also reduces jump, swim and other stamina drains when these items are equipped due to the point in the process of stamina reduction we are hooking it.

I don't think its worth merging this if we have not fixed that core problem yet.

AccretionCD:
I suppose the branch name is now a bit inappropriate considering the scope of the changes :smile:

I've tested all weapon types and tools, building, repairing and terrain manipulation while jumping, sprinting, etc.

For the tools patch, I've to have it run a stack trace to determine the calling method as the comment pointed out. I'm not sure how doing it in such a way would affect performance, please let me know if there are any issues.

Issue #154, opened 2021-02-26
Is the nightTimeSpeedMultiplier parameter not working?


muro1214:
I set the nightTimeSpeedMultiplier parameter as follows, but the night time does not seem to be accelerating.

~~~
; Total amount of time one complete day and night circle takes to complete
totalDayTimeInSeconds=1200

; Increase the speed at which time passes at night by %. The value 50 would result in a 50% reduced amount of night time.
nightTimeSpeedMultiplier=98
~~~

Maybe, the default case is...
* total day and night time = 1200 sec
* day time = 600 sec
* night time = 600 sec

nightTimeSpeedMultiplier=98 case is...
* total day and night time = 1200 sec
* day time = 600 + 600 x 0.98 = 1188 sec
* night time = 600 x ( 1- 0.98) = 12 sec

But it stayed night for at least five more minutes.

[Envaironment]
* Game version = 0.146.11
* V+ version = 0.9

JF10R:
Although it doesn't fix your issue, I would like to clarify the default values:
Day time is 21 minutes.
Night time is 9 minutes.

Source: https://valheim.fandom.com/wiki/Day_and_Night_Cycle

muro1214:
@JF10R Thank you for the information on default values!

xiaodoudou:
Base on the code, it seems it triggers after the **Awake** event.
Which could means it should trigger from my understanding after a first cycle?

Base on your configuration totalDayTimeInSeconds 1200s => 20 minutes
Base on my assumption the default value daytime is 55% and night time is 45% and we have a hook that triggers every time the time is updated on the server-side.

This means the night will be multiplied on each server tick for the updated time; in your case, you will make each tick going 98% faster. What it does that it bumps the delta of time between the previous and current tick and multiplies it by 98% not on his global aspect as the server tick will always occur on the same tick aspect.

So you wish that your nighttime to be 12 seconds, but it will take more time as it does that on tick aspects.

More tests need to be done, but I would suggest you go around 294 as a multiplier; without knowing what is the tick rate of that event the math will be a bit random to give you. My logic is if 98 takes ~5min; 130 should give you ~2.5min; 294 should give less than a minute.

Will give a try on my side to help you.

xiaodoudou:
From my point of view, to simplify the logic, to have more linear time calculation.
At the moment we do, time spent + time spent * multiplier

The default value should be 1.0 and we should act on the tick without adding it again.

The code is:
```   
[...]

Issue #171, opened 2021-02-27
Improve configuration note for Fireplace/onlyTorches


xiaodoudou:
Hi,

We should improve the configuration note for the onlyTorches parameter we can see here:
https://github.com/valheimPlus/ValheimPlus/blob/80831fc4a62805a1a18103b2ef8b0aa26859fd13/valheim_plus.cfg#L87

If the section is set to Flase, the onlyTorches parameter won't be triggered.

rr222cy:
Agreed, I suggest formatting it like:

```
[Fireplace]
needToRefillFireplaces = true;
needToRefillTorches = false;
```

Issue #170, opened 2021-02-27
Sleeping Issue


Graghtok:
We’ve got a strange issue since installing Valheim Plus where the sleep timer seems to be different than the time of day. When it’s clearly nighttime, our characters just lay in  bed endlessly. We did modify the day and night length parameters in the config. Has anyone else experienced this? Not seeing much information out there.

MiR4i:
I have not yet encountered that issue and I am altering the day/night cycle through V+, but just to be clear, every character currently playing on your server, regardless of their location in the world, is lying down at the same time?  In other words, if there are four active players on a server, then all four will need to be lying in a bed to "go to sleep" and pass through to the next day, otherwise your character(s) will just lie there and rest.

https://valheim.fandom.com/wiki/Day_and_Night_Cycle

While there are mods out there that can adjust the player threshold for sleep to occur and quickly pass through the night, V+ does not currently offer this option.

Xaphanie:
We're having the same issue, night / sleeping occurs a long time after it's actually dark and when you wake up it's way into the next day so something is off for sure.

xiaodoudou:
@Xaphanie did the time accelerated at least?

xiaodoudou:
I do reproduce the issue when you are very close to the daytime.

Graghtok:
Yes we've ensured that everyone is laying down at the same. 

Since sleeping isn't typically working for us, we're up all night in-game and the night cycles don't appear to be accelerated. Like Xaphanie noted, sometimes (when sleep does work) we wake up and it appears to be mid-day. 

Overall, not a not a deal breaker for us using the mod. Just a bit annoying to have to try to build during the middle of the night.

WAR-Lust:
I have the same issue playing on dedicated server a friend made. I know he has modified the day/time cycles but so far the issue encountered where (i use a Clock mod to have the ingame time):

- My game was middle of the night (2am) and his was day time 
- Playing solo on the server: While lying in bed, sleep activate at different time (usually 00:00 but sometimes 8pm, 11pm or 2am) 

When sleeping the time is sped up and depending on activation wake up is between 4am and 11am

babodad:
We are playing on a dedicated server and got the same issue. I did not change any of the time settings. If only one player is on the server sleeping does work. But with 2+ players everybody is just laying in their beds and nothing happens.

nxPublic:
Should be resolved in 0.9.3

PR #172, opened 2021-02-27
Update Time logic to be aligned with configuration + update config (#154/#171)

xiaodoudou:
Adjust time logic calculation to align configuration and solve reported bug #154


xiaodoudou:
Updated the config notes for #171

xiaodoudou:
People are complaining sometimes they cannot sleep.

I'm suspecting the following check to be in cause:

```cs
            if (!ZNet.instance.IsServer())
            {
                Debug.Log("Is not server.");
                return;
            }
```

My assumption is maybe the client sync with the server and both have their time sync periodically. 
What put be on that is the following report we got on discord:
```
Dat Rigby ThoToday at 9:58 AM
Looks like this might be a bug. Are you doing this on a dedi? I went and changed my client settings to match my server settings and now it's actually working. It seems like the server knows it's still "day" which is why I couldn't sleep until deep into the night. I went and changed my client settings and logged back in, and now it's daytime in the server instead of night.
Seems like there's some desync between the time of day on server and what the client settings are expecting.
Guess I'll be leaving it vanilla until that's fixed.
```

Maybe what we are having is a local client seen a night time when it's a day time on the server?
A test of 3 people could be a confirmation for it or more debugging and checking what the game is doing could confirm this.

xiaodoudou:
Just to link to that PR with bug #170; better cover the all in one.
I am running some tests related to the `isServer` condition to see if that what's troubling sleeping to be triggered.

xiaodoudou:
Here the result of testing:

- Environment: Dedicated Linux server with V+ 9.0.0 + Windows client on V+ 9.0.0
- Unusual behavior: Was able to sleep during day time
- The condition `Is not server.` is triggering only when a sleeping phase on the client side
- Configuration is:

```
[Time]

; Change false to true to enable this section
enabled=true

; Total amount of time one complete day and night circle takes to complete
totalDayTimeInSeconds=1800

; Increase the speed at which time passes at night by %. The value 50 would result in a 50% reduced amount of night time.
nightTimeSpeedMultiplier=50
```

xiaodoudou:
After some debug on client/server; seems the `Boolean isNight = __instance.IsNight();` not in phase with the sleeping feature.
Was able to sleep when the server was returning `isNight at false`

And after got weird desync on time when the client was lock in `isNight at true` and server was always considering its day time `isNight  at false`.

Trying to see how we can work around that.

xiaodoudou:
After checking Valheim code, It seems we are able to sleep during the afternoon:
```cs
[...]
else if (this.IsMine())
	{
		ZLog.Log("Is mine");
		if (this.IsCurrent())
		{
			ZLog.Log("is current spawnpoint");
			if (!EnvMan.instance.IsAfternoon() && !EnvMan.instance.IsNight())
			{
				human.Message(MessageHud.MessageType.Center, "$msg_cantsleep", 0, null);
				return false;
			}
[...]
```

Adding the time-skip check seems to solve the time desync.

xiaodoudou:
Ready for review, but that approach seems to works after my testing:

- Only the server is emitting netTime
- Avoid speed time during a time skip (Sleeping/Bedtime)
- Ensure we don't overlap the morning start (0.25% of a day)

I have tested it for a very short period and was able to see stuff working properly, note the ability to sleep during the afternoon is confusing at first, but it is intended by Valheim vanilla.

xiaodoudou:
forget to push my lastest commit.

ValheimPlus/RPC/VPlusConfigSync.cs
```diff
@@ -79,6 +79,9 @@ public static void RPC_VPlusConfigSync(long sender, ZPackage configPkg)
 
                             Configuration.Current = ConfigurationExtra.LoadFromIni(memStream);
 
+                            // Needed to make sure client is using server configuration as dayLength is setup before
+                            TimeManipulation.SetupDayLength();
```
xiaodoudou, line 83: @Mixone-FinallyHere not sure if I should put that here; but to highlight the issue; I have spotted if the client has different dayLength it is set up before network exchange, but doing such call here realign it.

xiaodoudou:
Merge time changes with 0.9.2.

If someone could give a test to see if it works fine and didn't forget anything that would make 2 bugs issues solved.

radmint:
@xiaodoudou could you merge master in so that all of the recent changes aren't showing as "Files changed" in the PR to make it a bit easier to read? Thank you!

xiaodoudou:
@radmint done ;-)

valheim_plus.cfg
```diff
@@ -110,7 +110,8 @@ fermenterItemsProduced=6
 ; If changed to enabled all fireplaces do not need to be refilled.
 enabled=false
 
-; If you enable this option only placed torches do not need to be refilled.
+; If you enable this option with the previous option to true, only placed torches do not need to be refilled. 
+; Apply to: wood torches, iron torches, green torches, sconces and brazier
```
radmint, line 114: thank you for this update!

nxPublic (approved):

nxPublic:
This does look like the correct solution, we will make sure to test and debug this before public release.

Awesome work!

❤️

PR #183, opened 2021-02-28
Add effects modifiers

xiaodoudou:
Small patch allowing to interact on modifier/multiplier/etc on the following Buff/Debuff:

- rested
- resting
- corpse run
- freezing
- cold
- wet


xiaodoudou:
When you do a review, please check the change done on ValheimPlus/Configurations/BaseConfig.cs
New update cover:

- Buff/Debuff percentage base to simplify the understanding of the setting
- Add boss buff hooks to change attackSkill, DamageTypeModifiers, etc...

ValheimPlus/Configurations/BaseConfig.cs
```diff
@@ -93,6 +93,12 @@ public void LoadIniData(KeyDataCollection data)
                     continue;
                 }
 
+                if (prop.PropertyType == typeof(string))
```
Mixone-FinallyHere, line 96: This currently clashes with the KeyCode property fetch, we are doing things bts to get this prop check to work

valheim_plus.cfg
```diff
@@ -366,4 +366,106 @@ enabled=false
...
+moderBuffStaminaRegenMultiplier = 0
+moderBuffJumpStaminaModifier = 0
+moderBuffRunStaminaModifier = 0
+moderBuffCooldown = 1200
+moderBuffDuration = 300
+moderBuffModifyAttackSkill = 
+moderBuffDamageModifier = 0
+moderBuffDamageTypesModifiers = 
+moderBuffDescription = 
+
```
Mixone-FinallyHere, line 471: This is bloating config file WAY too much

too many configuration items

Mixone-FinallyHere (commented):
I would not add this feature as it is, way too many config settings, we are working on improving the config system, adding such granularity is a bit much, not so much customization is needed in this mod. 
It might be better to make a separate mod that can be used in conjunction with V+ for this

ValheimPlus/Configurations/BaseConfig.cs
xiaodoudou: Not super happy with the configuration approach parsing of all these parameters, having arrays would be a nice addition, let's wait for your next milestone on how you improve your config systems.

valheim_plus.cfg
xiaodoudou: It can be separated into a dedicated plugin, but then it will allow players to have their own config not sync with the server.
This is the reason why I did that PR to have it inside V+ which handles that but I totally understand your points ;-).

I don't know if you have any plan to have sub-modules or any approach that would basically allow/reject other plugins on the connection to the server and allow their configuration exchanges.

Let me know your thought on that.

Mixone-FinallyHere: That is something to look into

part of the reason we have trouble with external plugins is being ale to enable/set them server side

best idea is likely to for all mods to implement simialr solution to ours where you get the settings from server, but this is more of a game limitation sadly

xiaodoudou: If you want me to simplify the effects list and settings, let me know.
I can have some time today to clean it and adjust the one you guys wants.

xiaodoudou:
@Mixone-FinallyHere please review my last commit to see if that suit a better configuration approach keeping your idea of reducing configuration

ValheimPlus/GameClasses/ObjectDB.cs
```diff
@@ -0,0 +1,249 @@
...
+                    HitData.DamageModPair damageModPair;
+                    damageModPair.m_type = ParseDamageType(damageTypeModifierConfiguration[0]);
+                    damageModPair.m_modifier = ParseDamageModifier(damageTypeModifierConfiguration[1]);
+                    damageTypesModifiersListing.Add(damageModPair);
+                }
+            }
+            return damageTypesModifiersListing;
+        }
+
+        private static void DisplayEffectStats(SE_Stats currentStatus, bool isBefore)
```
Mixone-FinallyHere, line 109: Can we have less debug prints ? 

The debug option makes sense in dev, not in release

valheim_plus.cfg
```diff
@@ -368,3 +368,43 @@ wagonBaseMass=20
 
 ; This value changes the game physical weight of Vagons by +/- more/less from item weight inside. The value 50 would increase the weight by 50% more. The value -100 would remove the entire extra weight.
 wagonExtraMassFromItems=0
+
+[Effects]
+
+; Change false to true to enable this section
+enabled=false
+
```
Mixone-FinallyHere, line 376: what is being enabled here? everything is commented out

Mixone-FinallyHere (changes requested):
Too much debug stuff, and not sure what you are doing with config, I only see the enabled

valheim_plus.cfg
xiaodoudou: Its commented, but you can enable buff changes with configuration such as:

```
; Wet Debuff reduction Example
wet.staminaRegenMultiplier = -5
```

xiaodoudou:
Base on the contributing rules; as this is unfinished; I will close it for now ;-).
It is unfinished at this point.

PR #191, opened 2021-02-28
Fixes and additions to the StaminaUsage section

AccretionCD:
Closed PR #160 as changes go beyond my original intentions.

Fixed the previous issues regarding stamina drain being reduced for all activities when affected weapons/tools are equipped; and added an option to adjust stamina drain for shields.
I've tested all weapon types and tools, building, repairing and terrain manipulation while jumping, sprinting, etc. so it should all be in order.

For the UseStamina patch, I've had to have it run a stack trace to determine the calling method as the comment pointed out. I'm not sure how doing it in such a way would affect performance, please let me know if there are any issues. 🙂


Mixone-FinallyHere (commented):
Looks good to me
@radmint want to double check?

ValheimPlus/PlayerStats.cs
```diff
@@ -29,89 +29,111 @@ private static void Postfix(Player __instance)
...
-                    case "Pickaxes":
-                        v = v - (v * (Configuration.Current.StaminaUsage.pickaxes) / 100);
-                        break;
-                    default:
-                        break;
+                    if (weaponType != "")
+                    {
+                        switch (weaponType)
+                        {
+                            case "Swords":
```
radmint, line 64: Can we use the `Skills.SkillType` enum for comparisons here for better future base game compatibility?

ValheimPlus/PlayerStats.cs
```diff
@@ -29,89 +29,111 @@ private static void Postfix(Player __instance)
...
-                if (isHammer)
+    [HarmonyPatch(typeof(Player), "UseStamina")]
+    public static class ChangeStaminaUsageOfTools
+    {
+        private static void Prefix(ref Player __instance, ref float v)
+        {
+            if (Configuration.Current.StaminaUsage.IsEnabled)
+            {
+                string methodName = new StackTrace().GetFrame(2).GetMethod().Name;
+                if (methodName == "UpdatePlacement" || methodName == "Repair")
```
radmint, line 108: instead of these strings, could we run something like `methodName.Equals(nameof(Player.UpdatePlacement))`?

radmint (changes requested):
Excellent use of calling method, thank you for describing how you tested. This looks great, just requesting small updates and I'd be happy to approve. Thank you for the improvement!

radmint:
Oh @AccretionCD and please add your credits to the README.md

AccretionCD:
Thank you for taking the time to review. I'll get right on the changes. 👍

ValheimPlus/PlayerStats.cs
AccretionCD: Done.

ValheimPlus/PlayerStats.cs
AccretionCD: Reviewing this section led me to discover an oversight with blocking which I fixed. Previously only affected blocking with a shield and not with any other weapon.
Thank you 😄

AccretionCD:
I've made the requested changes + a small fix. Should all be in order.

radmint (dismissed):
Looks great! Thank you!

radmint:
@AccretionCD just needs updating from our latest push and it's good to go! Thanks again!

AccretionCD:
Merged and tested as per usual. Should be good. 🙂

radmint:
@AccretionCD would you mind merging master in so that your PR doesn't show 49 files changed? Thank you!

AccretionCD:
@radmint Done. 👍

Mixone-FinallyHere (approved):
LGTM!

nxPublic:
Very well done!

zedle (approved):

zedle:
Looks good, merging.

PR #206, opened 2021-03-02
Exception fix on ConfigurationExtra/getBool

xiaodoudou:
A small fix that cast value to string to avoid exception when configuration key is missing from the INI file.


radmint (approved):
Thank you!

zedle (approved):

PR #212, opened 2021-03-03
add ZNet extensions for instance scope determination

sirskunkalot:
@zedle As requested. No namespace btw, so you would not have to import it to use the extension. Did not manage to test the last condition of having peers on your local instance.

### DIFF
--- a/ValheimPlus/Utility/ZNetExtensions.cs
+++ b/ValheimPlus/Utility/ZNetExtensions.cs
@@ -1,0 +1,33 @@
+﻿public static class ZNetExtensions
+{
+    public static bool IsLocalInstance(this ZNet znet)
+    {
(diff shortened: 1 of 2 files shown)


ValheimPlus/Utility/ZNetExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+    public static bool IsServerInstance(this ZNet znet)
+    {
+        return !znet.IsLocalInstance() && !znet.IsClientInstance();
+    }
+
+    public static string GetInstanceType(this ZNet znet)
+    {
+        if (znet.IsLocalInstance())
+        {
+            return "Local";
```
zedle, line 22: This seems like a perfect spot for an Enum - just define it in the same file under ZNetExtensions

zedle (changes requested):
Looks good, I just think an Enum is better for GetInstanceType

zedle (approved):

PR #216, opened 2021-03-03
fix sleeping issue

xiaodoudou:
To be more tested, but so far didn't get a sleep issue with such an approach.
Note you can sleep in the afternoon base on Valheim code; don't ask me why it is like that.

### DIFF
--- a/ValheimPlus/GameClasses/EnvMan.cs
+++ b/ValheimPlus/GameClasses/EnvMan.cs
@@ -101,7 +101,7 @@ namespace ValheimPlus
                 if (ZNet.instance.IsServer() && ZNet.instance.GetNrOfPlayers() > 0 && !__instance.IsTimeSkipping() && __instance.IsNight())
                 {
                     double timeSeconds = ZNet.instance.GetTimeSeconds();
-                    double num = timeSeconds + Helper.applyModifierValue(Time.deltaTime, Configuration.Current.Time.nightTimeSpeedMultiplier);
+                    double num = timeSeconds + Helper.applyModifierValue(Time.fixedDeltaTime, Configuration.Current.Time.nightTimeSpeedMultiplier);
                     ZNet.instance.SetNetTime(num);
                 }
             }
--- a/ValheimPlus/GameClasses/EnvMan.cs
+++ b/ValheimPlus/GameClasses/EnvMan.cs
@@ -92,7 +92,7 @@ namespace ValheimPlus
                 if (!Configuration.Current.Time.IsEnabled) {
                     return;
                 }
-                if (ZNet.instance.IsServer() && ZNet.instance.GetNrOfPlayers() > 0 && !__instance.IsTimeSkipping() && __instance.IsNight()) {
+                if (ZNet.instance != null && ZNet.instance.IsServer() && ZNet.instance.GetNrOfPlayers() > 0 && !__instance.IsTimeSkipping() && __instance.IsNight()) {
                     double timeSeconds = ZNet.instance.GetTimeSeconds();
                     double num = timeSeconds + Helper.applyModifierValue(Time.fixedDeltaTime, Configuration.Current.Time.nightTimeSpeedMultiplier);
                     ZNet.instance.SetNetTime(num);
--- a/ValheimPlus/GameClasses/EnvMan.cs
+++ b/ValheimPlus/GameClasses/EnvMan.cs
@@ -18,8 +18,16 @@ namespace ValheimPlus
                 if (EnvMan.instance)
                 {
                     EnvMan.instance.m_dayLengthSec = (long)Configuration.Current.Time.totalDayTimeInSeconds;
+                    Debug.Log($"Setup m_dayLengthSec: {EnvMan.instance.m_dayLengthSec}");
                 }
             }
+        }
+
+        public static double GetDayProgression()
+        {
+            double timeSecondsFromStart = ZNet.instance.GetTimeSeconds();
+            double timeSecondsFromDay = timeSecondsFromStart % Configuration.Current.Time.totalDayTimeInSeconds;
+            return timeSecondsFromDay / Configuration.Current.Time.totalDayTimeInSeconds;
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


xiaodoudou:
@nxPublic did run with several configs and doesn't encounter the issue:

- 240 totalDayTimeInSeconds+ 100% boost
- 1800 totalDayTimeInSeconds + 50% boost

To be tested by alpha tester.

ValheimPlus/GameClasses/EnvMan.cs
```diff
@@ -25,41 +25,91 @@ public static void SetupDayLength()
...
+        public static class EnvMan_Awake_Patch
         {
             private static void Prefix(ref EnvMan __instance)
             {
                 SetupDayLength();
             }
         }
 
+        [HarmonyPatch(typeof(EnvMan), "IsDay")]
+        class EnvMan_IsDay_Patch
```
radmint, line 55: 1. Can we add the class modifiers to follow (admittedly undocumented) convention? eg `public static` in front of `class`?
2. Could we add `///summary` above these new patches?

ValheimPlus/GameClasses/EnvMan.cs
```diff
@@ -25,41 +25,91 @@ public static void SetupDayLength()
...
+        {
+            static bool Postfix(bool result)
+            {
+                if (ZNet.instance.IsServer())
```
radmint, line 59: Do you think we should check if `ZNet.instance` isn't `null` before calling it?

ValheimPlus/GameClasses/EnvMan.cs
```diff
@@ -25,41 +25,91 @@ public static void SetupDayLength()
             }
         }
 
+        private const float startOfADay = 0.25f;
```
radmint, line 28: Where did we get these values from?

ValheimPlus/GameClasses/EnvMan.cs
```diff
@@ -25,41 +25,91 @@ public static void SetupDayLength()
...
         /// Hook on EnvMan to alter night speed
         /// </summary>
         [HarmonyPatch(typeof(EnvMan), "FixedUpdate")]
-        public static class TimeUpdateHook
+        public static class EnvMan_FixedUpdate_Patch
         {
             private static void Prefix(ref EnvMan __instance)
             {
-                if (Configuration.Current.Time.IsEnabled)
+                if (!Configuration.Current.Time.IsEnabled)
```
radmint, line 104: Do we need to add this check in front of the other patches above as well?

radmint (changes requested):
Thanks @xiaodoudou! Just some small updates requested. Could you also include your test cases in your PR summary and give a brief description of how your changes fix the issue?

ValheimPlus/GameClasses/EnvMan.cs
xiaodoudou: will do

ValheimPlus/GameClasses/EnvMan.cs
xiaodoudou: Lets do that just in case; didnt got any crash but worth avoiding it

ValheimPlus/GameClasses/EnvMan.cs
xiaodoudou: Its "magic number" wrote on the original methods, and they don't have constant :-/

ValheimPlus/GameClasses/EnvMan.cs
xiaodoudou: Let me add that

radmint (dismissed):
This looks good to me. I'll grab another review from another dev and we'll get this tested. Thanks @xiaodoudou!

ValheimPlus/GameClasses/EnvMan.cs
radmint: All good just curious!

xiaodoudou:
@radmint found one condition where `ZNet.instance != null` was missing as you guys did request I do on all of it.
Will commit that just for having that check ;-).

Did run some test today, my world is 2300+ days and did succeed to make the sleep triggered and see the night speed up.
We need to run a test with more than one player to ensure it is working properly. 
My test was always on:
 - Local Server with me alone
 - Linux Server with me alone on the server

Clarification to the tester: 
- Valheim Vanilla allow sleep during the afternoon; this not a bug but a feature ;-)
- If you can lie down on the bed; it means the local client considers its either afternoon or night, if all players connected are attached to a bed, it will check on the server-side periodically if it's the afternoon or night, then that's the case will trigger TimeSkip (Sleeping)

Here the code calls for clarification
```cs
	private void UpdateSleeping() // Call periodically
	{
		if (!ZNet.instance.IsServer())
		{
			return;
		}
		if (this.m_sleeping)
		{
			if (!EnvMan.instance.IsTimeSkipping())
			{
				this.m_sleeping = false;
				ZRoutedRpc.instance.InvokeRoutedRPC(ZRoutedRpc.Everybody, "SleepStop", Array.Empty<object>());
				return;
			}
		}
		else if (!EnvMan.instance.IsTimeSkipping())
		{
			if (!EnvMan.instance.IsAfternoon() && !EnvMan.instance.IsNight())
			{
				return;
			}
			if (!this.EverybodyIsTryingToSleep())
			{
				return;
			}
			EnvMan.instance.SkipToMorning();
			this.m_sleeping = true;
			ZRoutedRpc.instance.InvokeRoutedRPC(ZRoutedRpc.Everybody, "SleepStart", Array.Empty<object>());
		}
	}

	private bool EverybodyIsTryingToSleep() // Call by UpdateSleeping() if IsAfternoon() OR IsNight() is true
	{
		List<ZDO> allCharacterZDOS = ZNet.instance.GetAllCharacterZDOS();
		if (allCharacterZDOS.Count == 0)
		{
			return false;
		}
		using (List<ZDO>.Enumerator enumerator = allCharacterZDOS.GetEnumerator())
		{
			while (enumerator.MoveNext())
			{
				if (!enumerator.Current.GetBool("inBed", false))
				{
					return false;
				}
			}
		}
		return true;
	}
```

The test plan example:
- Setup: 240 totalDayTimeInSeconds+ 100% boost
  - Player 1 and Player 2 wait 2 days to spend (to see if any desync occurs), during that test, they should see if night, day, afternoon are in sync with each other
  - Player 1 trigger sleep function in afternoon (will be able to lie down but no sleep triggered)
  - Player 2 trigger sleep function at the night (will be able to lie down and sleep should be triggered)
  - Both player woke up in the morning, and wait afternoon to trigger sleep (it should be triggered)
  - Both player woke up in the morning, and wait the night to trigger sleep (it should be triggered)

The idea to test:
- If any desync time is happening on different players (client-side)
- If any time calculation is wrong on the server-side which is a must-have for the sleep to trigger

xiaodoudou:
Closing this PR; Im planning to rework it base on RescaleDayFraction
Will re-open when ready and tested.

PR #218, opened 2021-03-03
Feature/free rotation with attach

WolfgardV:
Rotates placementGhost by 1 degree, if pressed key, or reset to usual rotation x22.5f degrees, if rotate without key pressed.
Unlike in your AdvancedBuildingMode, this variant still use auto-attachment to nearly objects.
Also, it saves rotation prom previous placement, so you can qickly place a lot of objects in line with un-standard rotation.
Also, added hot keys to copy rotation from target piece, and another to set rotation as perpendicular to selected piece.

Actually, you can attach to it yout Num+ and Num- scale functionality.

### DIFF
--- a/ValheimPlus/Configurations/Configuration.cs
+++ b/ValheimPlus/Configurations/Configuration.cs
@@ -31,5 +31,6 @@ namespace ValheimPlus.Configurations
         public CameraConfiguration Camera { get; set; }
         public GameConfiguration Game { get; set; }
         public VagonConfiguration Wagon { get; set; }
+        public FreePlacementRotationConfiguration FreePlacementRotation { get; set; }
     }
 }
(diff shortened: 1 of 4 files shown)


radmint:
@WolfgardV please fix merge conflicts and just a heads up, we changed head branch to development so pull in development branch to fix conflicts. We'll be able to review soon, this will just require a bit more time.

WolfgardV:
Updated PR to develop, fixed conflicts.

nxPublic (approved):
I don't see any issues, tests will show.

I would have really liked some more comments in it, even if its copied from original game code.

WolfgardV:
Yes, it contains original game code, and I don't like it a lot. but it was the only way, I found, to get around that harcoded 22.5 angle and sync auto-ttachment after apply my rotation (

PR #224, opened 2021-03-04
- Added Crop notifier when a player tries to place a crop too close t…

beerik94:
This adds an onscreen message notification that the player is trying to place a crop too close to another crop.

### DIFF
--- a/valheim_plus.cfg
+++ b/valheim_plus.cfg
@@ -248,6 +248,9 @@ disableCameraShake=false
 ; The base unarmed damage multiplied by your skill level
 baseUnarmedDamage=10
 
+; When changed to true, makes you unable to place crops too close to eachother
+cropNotifier=false
+
 [Server]
 
 ; Change false to true to enable this section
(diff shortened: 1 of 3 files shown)


valheim_plus.cfg
```diff
@@ -248,6 +248,9 @@ disableCameraShake=false
 ; The base unarmed damage multiplied by your skill level
 baseUnarmedDamage=10
 
+; When changed to true, makes you unable to place crops too close to eachother
+cropNotifier=false
```
radmint, line 252: I would put a bit more clear wording on the description. Maybe: "When changed to true, you will not be permitted to place a crop that is considered to close to another crop"

radmint (changes requested):
Thank you for the contribution!

1. Would you mind updating your summary to include what test cases you performed with this?
2. Did you try this feature in conjunction with the "no invalid placement" V+ mod under the [Building] section?
3. Please add a comment in your code addition to describe what it's doing

beerik94:
Added comments and changed the config file.
This works even when you have "No invalid placement" turned ON.
I don't know if you want to keep it like that.
If not then I will have to look into fixing it.

radmint (approved):
These updates look great thank you @beerik94! Going to get a second set of eyes and test locally but then it should be ready to merge!

nxPublic:
Awesome work and solid implementation!

Thank you! ❤️

nxPublic (approved):

PR #247, opened 2021-03-05
simplifying ZNetExtensions

sirskunkalot:
I did some more and thorough testing, the constants in ZNet are sufficient. This is still a nice shortcut and if the evaluation rules should change, it means just changing this extension

### DIFF
--- a/ValheimPlus/Utility/ZNetExtensions.cs
+++ b/ValheimPlus/Utility/ZNetExtensions.cs
@@ -6,20 +6,20 @@
         Client,
         Server
     }
-     
+
     public static bool IsLocalInstance(this ZNet znet)
     {
-        return !znet.IsDedicated() && znet.IsServer() && (znet.GetNrOfPlayers() > znet.GetPeers().Count);
+        return znet.IsServer() && !znet.IsDedicated();
     }
 
     public static bool IsClientInstance(this ZNet znet)
     {
-        return !znet.IsDedicated() && !znet.IsServer();
+        return !znet.IsServer() && !znet.IsDedicated();
     }
 
     public static bool IsServerInstance(this ZNet znet)
     {
-        return !znet.IsLocalInstance() && !znet.IsClientInstance();
+        return znet.IsServer() && znet.IsDedicated();
     }
 
     public static ZNetInstanceType GetInstanceType(this ZNet znet)
@@ -27,7 +27,7 @@
         if (znet.IsLocalInstance())
         {
             return ZNetInstanceType.Local;
-        } 
+        }
         else if (znet.IsClientInstance())
         {
             return ZNetInstanceType.Client;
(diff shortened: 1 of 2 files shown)


ValheimPlus/Utility/ZNetExtensions.cs
```diff
@@ -6,28 +6,28 @@ public enum ZNetInstanceType
         Client,
         Server
     }
-     
+
     public static bool IsLocalInstance(this ZNet znet)
```
zedle, line 10: So this is a local server?

sirskunkalot: Yes, local instance is the local server with someone playing the game. But feel free to change the terminology as you like. I am doing my own mod for me and my friends, so I just contribute what I think might benefit you.

ValheimPlus/Utility/ZNetExtensions.cs
```diff
@@ -6,28 +6,28 @@ public enum ZNetInstanceType
...
     public static bool IsClientInstance(this ZNet znet)
     {
-        return !znet.IsDedicated() && !znet.IsServer();
+        return !znet.IsServer() && !znet.IsDedicated();
     }
 
     public static bool IsServerInstance(this ZNet znet)
     {
-        return !znet.IsLocalInstance() && !znet.IsClientInstance();
+        return znet.IsServer() && znet.IsDedicated();
```
radmint, line 22: @sirskunkalot do you know how `IsDedicated()` works? I see in base game, we just return `false` for that value so I'm wondering if we're overriding that somewhere that I'm not seeing.

sirskunkalot: It is a simple getter for a bool member of ZNet: ZNet.m_dedicated. This is true on all dedicated server builds. At least on my linux box this is the case, I did not install the Windows version yet.

> we just return false for that

Who returns that? V+? The vanilla ZNet class handles this correctly. Dedicated builds have `(m_server == true && m_dedicated == true)`. "Locally" started worlds (does not matter if "Start Server" is checked when starting the world) have `(m_server == true && m_dedicated == false)` even though you are also a player (but not a peer) at the same time and a local game, who "Joined Server" always has `(m_server == false && m_dedicated == false)`. I just shortcut it with my Extension and use the getters instead of the members. My previous version which tested NrOfPlayers() > NrOfPeers() is not necessary anymore but I did not want to rewrite my code again.

Don't know if this is the information you wanted, but HTH ;)

sirskunkalot: OK, now I got curious (and a little embarrassed how "thorough" my testing actually was) and installed the win dedicated on my work PC. 

valheim_win_dedicated

The marked output comes from this ZNet Extension. It says "Server". So yeah, unix and win dedicated servers always have `ZNet.m_dedicated` set to `true`.

radmint: @sirskunkalot and I figured out where we were seeing different results and it's because I was looking at the game.dll and he was looking at the dedicated game .dll. This all looks good to me!

radmint (approved):

Issue #231, opened 2021-03-05
[FEATURE REQUEST] Keep the mouse cursor in the same spot as when it looses focus


ChristopherHaws:
**Is your feature request related to a problem? If so, please describe.**
Every time I open and close chests, stop looking at my inventory, or exit the build menu my mouse cursor moves gets re-centered by the game so that when I reopen the chest, inventory, or build menu, I have to move my mouse back to the previous location.

**Describe the solution you'd like**
I would love it if the game remembered my mouse location when it looses focus instead of being re-centered on the screen.

**Additional context**
This is how Minecraft (at least bedrock edition) handles this and it feels very natural.

radmint:
![](https://github.trello.services/images/mini-trello-icon.png) [Keep the mouse cursor in the same spot as when it looses focus](https://trello.com/c/uFXPDfnq/55-keep-the-mouse-cursor-in-the-same-spot-as-when-it-looses-focus)

PR #248, opened 2021-03-06
Add the ability to remember the mouse position when it loses focus

ChristopherHaws:
Closes https://github.com/valheimPlus/ValheimPlus/issues/231

This feature is setup to only work on Windows since Unity doesn't have a way to set the mouse position out of the box.


zedle (approved):
Reasonable solution

nxPublic:
Good work, we will have to see how all the virus detection software is going to like the implementation of the Win32 Api though.
From my past experience that might cause a lot of false positives but I am willing to find out.

PR #251, opened 2021-03-06
Bug/config sync

xiaodoudou:
Hi,

First of all, I wish to apologize for my last commit on my previous PR that introduce the feature to simply don't work; it was a refactor base on a request; I did it too fast  and didn't cover a full test to see a simple assignment haven't been missed.
Again, sorry for that.

Anyway, I have reworked it:
- Revert the revert (just to be able to fix the code)
- Fix the previous PR that was introducing the bug that remote config to be parsed but not assigned
- Fix the previous PR logic that was considered a disable section from the server should be not parsed locally; now is aligned with the project logic/idea that if the server-config section is disabled; it will disable it on client-side too
- Make server not pushing KeyCode (useless data) and local client to load KeyCode from only local config on his logic
- Make config load keyCode even if section disabled; **please confirm** as it is to cover a sever enabling ABM/AEM and allowing client to configure his keyCodes with his local section enabled or disabled
- Adjust ABM to be server config section; **please confirm** if you wish that change
- Adjust AEM to be server config section; **please confirm** if you wish that change
- Adjust Workbench config section to be server config (We missed that one)
- Cleaner approach for default values for all kind of settings
- Improve logs for config; **please review** I can mute some of them but it helps to see the sequences of the configuration locally/sync from server
- If server-side disable a section it will reload the default value from the Class itself on local client (to avoid inconsistent configuration)

Please **be aware** that if we patch on a core Valheim classes Awake method (i.e. EnvMan, ObjectDB); they can be init before our config sync process, this nothing related to the sync feature itself but the chainload / Patch of such classes that is in place. The solution is either to patch a more common method/call or use a configure approach that is called after the client has successful process the remote configuration (i.e. Time patch).

Cheers,


ValheimPlus/Configurations/ConfigurationExtra.cs
```diff
@@ -54,49 +54,59 @@ public static bool LoadSettings()
...
+        }
+
         public static Configuration LoadFromIni(Stream iniStream)
+        {
+            using (StreamReader iniReader = new StreamReader(iniStream))
+            {
+                FileIniDataParser parser = new FileIniDataParser();
+                return LoadFromIni(parser.ReadData(iniReader));
+            }
+        }
```
zedle, line 88: Whitespace between methods

zedle:
These files need comments, following the code is quite difficult without any

ValheimPlus/Configurations/ConfigurationExtra.cs
xiaodoudou: will add it before `public static void LoadConfigurationFromStream(Stream iniStream)`

xiaodoudou:
@zedle did cover a code review by:
- adding comments
- fix some code style
- remove an always true condition

I have also found out that `RPC_VPlusConfigSync` is always doing down synchronization if `Server.serverSyncsConfig` is true regardless of the state of `Server.IsEnabled`. It is now covered.

xiaodoudou:
@radmint did handle the merge conflict.

PR #255, opened 2021-03-06
Bug/sleep fix

xiaodoudou:
Hi,

Here the sleep feature refactored:
- Patch as Prefix `EnvMan/RescaleDayFraction` instead of recreating server-side methods for: IsDay, IsAfternoon, IsNight
- Applying daysLength on `EnvMan/FixedUpdate` instead of `EnvMan/Awake`; it avoids resyncing the `m_dayLengthSec` after client been synchronized
- Make config be a full day and night cycle (by assigning `m_dayLengthSec` 2/3 of the value of `totalDayTimeInSeconds`)
- Ensure that if the Time section is disabled `m_dayLengthSec` is assigned to the standard value of Valheim Vanilla

There is one case not covered and needs alignment: 
- Server using V+ with Time section `enabled`
- Server section property `enforceMod` is set to `false`
- Client is not using V+

Result:
- Server will keep boosting the nights
- Client will be out of sync on the phase day time/night time

Cheers,

### DIFF
--- a/ValheimPlus/Configurations/Sections/TimeConfiguration.cs
+++ b/ValheimPlus/Configurations/Sections/TimeConfiguration.cs
@@ -2,7 +2,7 @@
 {
     public class TimeConfiguration : ServerSyncConfig<TimeConfiguration>
     {
-        public float totalDayTimeInSeconds { get; internal set; } = 1200;
+        public float totalDayTimeInSeconds { get; internal set; } = 1800;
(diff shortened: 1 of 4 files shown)


xiaodoudou:
@nxPublic please let me know you're though on making such feature linked to enforceMod.
Cheers,

nxPublic:
We have made the decision to not tamper with time for now. This brings to many potential issues with to little gain.

Issue #261, opened 2021-03-06
[BUG] Iron chest row setting is not obeyed. 


Sikareus:
- OS: [Windows Client/Windows Server]
- Valheim Game Version: [Latest]
 - Mod Version [0.9.4]

**Describe the bug**
The iron chest row config is not obeyed down to the minimum level. Even if you set the value to 3 rows, you will always have a minimum of 4 anytime the [Inventory] section is configured on using the 'true' flag.

**To Reproduce**
1. Set the server config to 3 rows and check an iron chest.

**Expected behavior**
You would expect the minimum row value to be obeyed.

radmint:
@Legionarius this is not a  bug, we clamp the minimum to 4 rows. If you would like the ability to go lower, please open a feature request.

Legionarius:
My mistake; was going by the comments in the config file. Perhaps just a super minor config file tweak for the future then?

; Iron chest number of rows (more than 4 rows will add a scrollbar)
; This also affects the cart and boat inventory
; min 3, max 20

radmint:
@Legionarius I found the issue, sorry about that I didn't realize our cfg and actual clamp was out of sync. Check out the PR that I just linked for a fix.

radmint:
Issue resolved, thank you again!

PR #262, opened 2021-03-06
Updating cfg for clarification, fixing min and max for iron chest

radmint:
Addresses #261 and syncs cfg with inventory col/row clamps


Mixone-FinallyHere (changes requested):
These changes will break the feature

The values changed were the mximum and minimums

Those are not to be changed

Configs are checked against max and min

Mixone-FinallyHere:
Ignore latest

I misread the MaxCol as MaxRow

Mixone-FinallyHere (approved):

radmint:
@Mixone-FinallyHere thanks for checking it out I was about to dm and ask about it haha

PR #266, opened 2021-03-06
Update maximumWood configuration to reflect the games default value

ChristopherHaws:
The default setting for the Kiln's maximumWood is 25 in the game, but the config file had it set to 20. The setting is correct in the C# config file however:
https://github.com/valheimPlus/ValheimPlus/blob/development/ValheimPlus/Configurations/Sections/KilnConfiguration.cs#L6


radmint (approved):
Confirmed that "max ore" on kiln prefab is 25.

PR #277, opened 2021-03-07
Add Debug libs and instructions

sirskunkalot:
The manual way, no VS debugging.


CONTRIBUTING.md
```diff
@@ -16,6 +16,18 @@ How to setup the development enviroment to compile ValheimPlus yourself.
...
 
+## Debugging with dnSpy
+
+Thanks to mono and unity-mono being open source, we patched and compiled our own mono runtime and enabled actual live debugging of the game and the mod itself with dnSpy.
+
+1. Download [dnSpy-net-win64](https://github.com/dnSpy/dnSpy/releases) and extract the exe.
+2. Load all assemblies from \<Valheim>\unstripped_corlib into dnSpy (just drag&drop the folder onto it).
+3. Load all assembly_* from \<Valheim>\valheim_Data\Managed into dnSpy (*do not load the publicized ones, they will not be loaded into the process and therefore can not be debugged*).
+4. Load ValheimPlus.dll from \<Valheim>\BepInEx\plugins into dnSpy.
+5. Copy .\libraries\Debug\mono-2.0-bdwgc.dll from this repo to \<Valheim>\MonoBleedingEdge\EmbedRuntime and overwrite the existing file.
```
Dfluke, line 27: You use relative files paths in all prior steps, but step 5 refers to a 'this' directory. I think updating this to a more verbose directory would be a solid idea for 2 reasons.
1. It removes any potential confusion about where 'this' is, and makes steps more beginner friendly.
2. In the event that the CONTRIBUTING.md file ever moves location within the project, the location referenced by 'this' would be wrong.

Dfluke, line 27: Rather than 'from this repo', do we have a simpler way to refer to 'this' (being the V+ checkout)?

Any thoughts on a var that refers to the project checkout location itself?

sirskunkalot: I honestly don't care. This is addressing developers, so if they don't understand, they should not be here in the first place.

sirskunkalot:
@radmint documentation? I, Madam, feel insulted. Documenting was the nice-to-have here ;)

CONTRIBUTING.md
Dfluke: How very elitist of you.

It wasn't about a lack of understanding, my thought was that as the project, and instructions grow, use of an abstract 'this' isn't in family with the other relative path. 

An idea to be mulled over. Not an action.

Resolved.

radmint: @sirskunkalot any time someone suggests clarification it should be taken into consideration. It could be misunderstood for more reasons than someone not being a dev. Miscommunication happens because of context loss, practicing english as a second language, etc. Please be more considerate of feedback on your PRs. I'm not saying implement anything that anyone asks but I don't think "devs would understand" is a valid reason to dismiss feedback.

Dfluke:
Agreed. Which is why it was just a comment tossed out as a "have we thought about"? 

Wasn't pressing you to make any changes. I'm excited to get home and try this :)

CONTRIBUTING.md
sirskunkalot: This PR is about the library, not the inclusiveness of the instructions. Change the wording, if you like,

sirskunkalot: > It wasn't about a lack of understanding, my thought was that as the project, and instructions grow, use of an abstract 'this' isn't in family with the other relative path.
Thank you for expressing your considerations. But after a reevaluation of the facts given, 'this repo' seems to me as the most precise description, as you don't need to get this particular file from a local clone necessarily. In contrast to the necessity of having the source files and the game files in a local environment to exercise the task at hand, one must not feel persuaded and forced by the instructions provided.

> How very elitist of you.
Oh and thanks for the insult.

Dfluke: Anytime :)

radmint (approved):

PR #279, opened 2021-03-07
alphabetized congif category. also format consistenly to enhance read…

ahnguyen09:
At the start of the config file, it was alphabetized at some the beginning then not. I just reorder the config so it is in alphabetical order. Then I format it to be consistent and more readable.

[Beehive]  <!no new line>
; Change false to true to enable this section
enabled=false

; configure the speed at which the bees produce honey in seconds, 1200 seconds are 24 ingame hours
honeyProductionSpeed=1200

; configure the maximum amount of honey in beehives
maximumHoneyPerBeehive=4  <!two new line>

[Building]

### DIFF
--- a/valheim_plus.cfg
+++ b/valheim_plus.cfg
@@ -16,8 +16,8 @@ copyObjectRotation=Keypad7
 ; Apply the copied object rotation to the currently selected object in ABM
 pasteObjectRotation=Keypad8
 
-[AdvancedEditingMode]
 
+[AdvancedEditingMode]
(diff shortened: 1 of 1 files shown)


radmint:
Thank you @ahnguyen09 we really appreciate this but we've stepped away from attempting to alphabetize because inserting new configurations into the middle of the .cfg file messes up how many people try to merge old and new cfgs together and/or parse them. It's definitely not ideal to have to ctrl+f to search for a configuration to change but we're working on a better system that hopefully won't involve editing the .cfg file directly.

PR #284, opened 2021-03-08
Monster Loot Delay

HeberPcL:
Guys, I added an option to control the time for the monster's loot to appear.

The Monster class already paves the way to centralize other related operations.

I'm new here, let me know if I need to adjust anything.


radmint (changes requested):
I'm not quite sure what this configuration change accomplishes. What's the drive behind a monster loot delay?

1. Please follow contributing guidelines in CONTRIBUTING.md for standards and conventions for our codebase
2. This configuration is missing from the valheim_plus.cfg in your PR
3. You aren't checking for IsEnabled for an overall configuration (see how other configs do it)
4. You haven't specified your test cases in your PR summary

HeberPcL:
The idea is to be able to control the time of death until the loot appears, this setting will reduce this delay.

The test to be done is to check if the speed between death and the appearance of the loot is less. Success in my tests.

Thank you for your considerations.

nxPublic (changes requested):

HeberPcL:
I updated the default value in class and cfg.

ValheimPlus/GameClasses/Monster.cs
```diff
@@ -0,0 +1,23 @@
...
+using ValheimPlus.Configurations;
+
+namespace ValheimPlus.GameClasses
+{
+    class Monster
+    {
+        /// <summary>
+        /// Altering time for loot appear
+        /// </summary>
+        [HarmonyPatch(typeof(Ragdoll), "Awake")]
```
nxPublic, line 11: If the affected class is Ragdoll, the class name of the file should be Ragdoll.cs

nxPublic (changes requested):

radmint:
@HeberPcL would you mind updating your summary with the test cases that you performed to test this?

valheim_plus.cfg
```diff
@@ -572,6 +572,14 @@ enabled=false
...
 ; Increase or decrease the block value on all shields in %. -50 would be 50% less block rating, 50 would be 50% more block rating.
 blockRating=0
+
+[Deconstruct]
+; Change false to true to enable this section
+enabled=false
+
+; Percentage of resource to return when deconstructing an item. The value 100 returns back 100% of resources required to make the item.
+percentageOfReturnedResource=100
+blockRating=0
```
nxPublic, line 582: Merge conflict slipped by.

nxPublic (changes requested):

HeberPcL:
> @HeberPcL would you mind updating your summary with the test cases that you performed to test this? Also since we're applying this the whole ragdall class, I'm not sure this will just affect monster drops. Because of this, we should consider renaming the section/configuration from using "Monster" or add specification on ragdoll

The only scenario I can imagine, is to kill any monster and check the delay for the loot. That was my test.

valheim_plus.cfg
```diff
@@ -579,3 +579,8 @@ enabled=false
 
 ; Percentage of resource to return when deconstructing an item. The value 100 returns back 100% of resources required to make the item.
 percentageOfReturnedResource=100
+blockRating=0
+
+[Monster]
+; Time of death until the loot appears ( default=0.05f )
+dropDelay=0.05f
```
nxPublic, line 586: Remove the f at the end here

nxPublic (changes requested):

HeberPcL:
I made a mod for this.

PR #286, opened 2021-03-08
Addressing cfg "opposite enforceMod" bug

radmint:
Found a duplicate version string check and moved version string patch to use Postfix instead of Prefix


xiaodoudou:
@radmint worth editing the config to inform users enforceMod will require that section to be enabled with it to enabled if a server is enforcing the mod.

radmint:
@xiaodoudou having the section `enabled=true` is the default for the whole configuration file and we have the note `; Change false to true to enable this section` above it.

nxPublic (approved):

PR #294, opened 2021-03-09
Properly apply radius increases of the CraftingStation and Ward to th…

drewaburden:
…eir EffectArea.

CraftingStation and Wards have a child GameObject that has an EffectArea and a collider. The mob spawning system uses this collision to check whether spawning is allowed, and the collision radius is not automatically updated with the CraftingStation/Ward radii. So, without this, mobs can still spawn inside those expanded areas. By adjusting the EffectArea as well, we can ensure that any modification to the ranges there apply to mob spawning as well.

### DIFF
--- a/ValheimPlus/GameClasses/CraftingStation.cs
+++ b/ValheimPlus/GameClasses/CraftingStation.cs
@@ -12,7 +12,7 @@ namespace ValheimPlus
         [HarmonyPatch(typeof(CraftingStation), "Start")]
         public static class WorkbenchRangeIncrease
         {
-            private static void Prefix(ref float ___m_rangeBuild, GameObject ___m_areaMarker)
+            private static void Prefix(CraftingStation __instance, ref float ___m_rangeBuild, GameObject ___m_areaMarker)
             {
                 if (Configuration.Current.Workbench.IsEnabled && Configuration.Current.Workbench.workbenchRange > 0) 
                 {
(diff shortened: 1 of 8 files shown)


ValheimPlus/GameClasses/CraftingStation.cs
```diff
@@ -12,7 +12,7 @@ class CraftingStationChanges
         [HarmonyPatch(typeof(CraftingStation), "Start")]
         public static class WorkbenchRangeIncrease
         {
-            private static void Prefix(ref float ___m_rangeBuild, GameObject ___m_areaMarker)
+            private static void Prefix(CraftingStation __instance, ref float ___m_rangeBuild, GameObject ___m_areaMarker)
```
nxPublic, line 15: You need to use "ref CraftingStation __instance"

nxPublic (changes requested):

ValheimPlus/GameClasses/CraftingStation.cs
drewaburden: The ref keyword is not needed here, and in fact it's not needed in several other places it is currently being used. Classes in C# are always passed by reference. Adding ref makes it a reference to the reference, which would only be useful if you wanted to replace the caller's representation of the instance itself.

drewaburden: But, to be clear, it is still needed on primitive-typed arguments.

drewaburden: In C++syntax, it would effectively be the difference between these:
`void Prefix(CraftingStation** __instance)` (with ref)
`void Prefix(CraftingStation* __instance)` (without ref)
but in C#, there's obviously no `->` operator, so the `.` operator always dereferences all pointers.

Needing the ref keyword on classes is a common misconception in C#, and in my opinion the language needs to be altered to be more explicit.

drewaburden:
Bleh. It's been ages since I've used Git or GitHub, and I think I've botched this pull request by adding on an unrelated changelist. I'm going to close it and make a new one.

PR #300, opened 2021-03-09
Added a number of new build-related features.

cstamford:
* Workbench attachment radius
* AoE repair
* Pieces always drop full and/or excluded resources when destroyed

These are from https://www.nexusmods.com/valheim/mods/53, integrating these brings V+ up to feature parity with that mod :)

I have tried to follow the contributing guidelines but this is my first PR so I probably screwed something up!

### DIFF
--- a/valheim_plus.cfg
+++ b/valheim_plus.cfg
@@ -66,6 +66,18 @@ noWeatherDamage=false
 ; The maximum range that you can place build objects at
 maximumPlacementDistance=5
 
+; When destroying a building piece, setting this to true will ensure it always drops full resources.
+alwaysDropResources=false
+
+; When destroying a building piece, setting this to true will ensure it always drops pieces that the devs have marked as "do not drop".
+alwaysDropExcludedResources=false
+
+; Setting this to true will cause repairing with the hammer to repair in a radius.
+enableAreaRepair=false
+
+; Tweaks the area repair radius
+areaRepairRadius=7.5
+
 [Camera]
 
 ; Change false to true to enable this section
@@ -457,6 +469,9 @@ enabled=false
 
 ; Set the workbench radius in meters
 workbenchRange=20
+
+; Sets the workbench attachment (e.g. anvil) radius
+workbenchAttachmentRange=5
 
 ; Disables the roof and exposure requirement to use a workbench
 disableRoofCheck=false
--- a/valheim_plus.cfg
+++ b/valheim_plus.cfg
@@ -67,15 +67,17 @@ noWeatherDamage=false
 maximumPlacementDistance=5
 
 ; When destroying a building piece, setting this to true will ensure it always drops full resources.
+; default: false
 alwaysDropResources=false
 
 ; When destroying a building piece, setting this to true will ensure it always drops pieces that the devs have marked as "do not drop".
+; default: false
 alwaysDropExcludedResources=false
 
 ; Setting this to true will cause repairing with the hammer to repair in a radius.
 enableAreaRepair=false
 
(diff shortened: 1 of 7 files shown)


valheim_plus.cfg
```diff
@@ -66,6 +66,18 @@ noWeatherDamage=false
 ; The maximum range that you can place build objects at
 maximumPlacementDistance=5
 
+; When destroying a building piece, setting this to true will ensure it always drops full resources.
```
radmint, line 69: Would you be able to add a line `; default: false` etc for these configurations? Not technically a convention we have listed but more a preference

valheim_plus.cfg
```diff
@@ -66,6 +66,18 @@ noWeatherDamage=false
...
+alwaysDropResources=false
+
+; When destroying a building piece, setting this to true will ensure it always drops pieces that the devs have marked as "do not drop".
```
radmint, line 72: I wonder if we should add some examples of some pieces that are considered "do not drop"?

valheim_plus.cfg
```diff
@@ -66,6 +66,18 @@ noWeatherDamage=false
...
+alwaysDropExcludedResources=false
+
+; Setting this to true will cause repairing with the hammer to repair in a radius.
+enableAreaRepair=false
+
+; Tweaks the area repair radius
```
radmint, line 78: Could we be more specific here? Maybe something like "Sets the area repair radius. A value of 7.5 would mean your repair radius is 7.5m. Requires `enableAreaRepair=true`"

radmint (changes requested):
Looks great! I'm not able to comment much on the IL code since I've just recently started diving into that myself but thank you so much for following our contributing conventions. I just have a couple of requests on .cfg comment clarifcations. Would you also be able to add some info about these features to the README.md?

valheim_plus.cfg
cstamford: I actually don't know exactly what things are marked as "do not drop". For some context, both of these options were created because I often found, when playing online with my partner, that sometimes destroying buildings would not refund all of the materials. I know a combination of both has eliminated that problem for us. If you/anyone has a good method of scanning the game data, looking for the field `m_recover` in any piece requirements, we could generate a list...

cstamford:
> Would you also be able to add some info about these features to the README.md

My partner (@NariKat) is going to help me with documentation. Is it OK if I submit she submits a PR after all of the existing PRs have been merged with the documentation for all of the changes, or would you prefer one-by-one?

radmint:
@cstamford yeah that's fine as long as it gets in before this work is included in a release.

valheim_plus.cfg
radmint: @Mixone-FinallyHere might know of a way to do this but for now it's fine as is

radmint (dismissed):

valheim_plus.cfg
```diff
@@ -458,6 +472,9 @@ enabled=false
 ; Set the workbench radius in meters
 workbenchRange=20
 
+; Sets the workbench attachment (e.g. anvil) radius
+workbenchAttachmentRange=5
```
nxPublic, line 476: I think the [Workbench] section would be better suited

nxPublic (commented):
If you change the config section to the Workbench, this is ready for merge 👍

valheim_plus.cfg
nxPublic: My bad. GitHub display tricked me 😓

nxPublic (approved):

PR #301, opened 2021-03-09
Food changes.

cstamford:
* Add option to disable food degradation.
* Refactor the code that handles food duration scaling.

This change rolls in PreWorkout from https://www.nexusmods.com/valheim/mods/54.

The reason I have refactored the existing food scaling code:

1. It returns false from prefix, meaning that is a compatibility nightmare.
2. It has the original function reimplemented, which will cause issues further down the line when the Valheim devs inevitably change the base game.

The refactored code is the minimal change to produce the same result and will incur significantly lower maintenance costs going forwards IMO.

### DIFF
--- a/ValheimPlus/GameClasses/Player.cs
+++ b/ValheimPlus/GameClasses/Player.cs
@@ -199,7 +199,7 @@ namespace ValheimPlus
 
             List<CodeInstruction> il = instructions.ToList();
 
-            if (!Configuration.Current.Food.foodDegradesOverTime)
+            if (Configuration.Current.Food.disableFoodDegradation)
             {
                 for (int i = 0; i < il.Count; ++i)
                 {
--- a/ValheimPlus/GameClasses/Player.cs
+++ b/ValheimPlus/GameClasses/Player.cs
@@ -1,6 +1,10 @@
 ﻿using HarmonyLib;
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
+using System.Reflection;
+using System.Reflection.Emit;
 using UnityEngine;
 using ValheimPlus.Configurations;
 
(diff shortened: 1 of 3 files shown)


ValheimPlus/GameClasses/Player.cs
```diff
@@ -135,77 +139,83 @@ private static void Postfix(ref Player __instance, ref Vector3 ___m_moveDir, ref
...
+        /// <summary>
+        /// Replaces loads to the current health/stamina for food with loads to the original health/stamina for food
+        /// inside Player::GetTotalFoodValue. This disables food degradation.
+        /// </summary>
+        [HarmonyTranspiler]
+        public static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstruction> instructions)
+        {
+            if (!Configuration.Current.Food.IsEnabled) return instructions;
+
+            List<CodeInstruction> il = instructions.ToList();
```
radmint, line 200: Do you think this method would be a good candidate for the `yield return` concept that's recommended for transpilers?

valheim_plus.cfg
```diff
@@ -140,6 +140,9 @@ enabled=false
 ; Allows for - values to shorten food duration.
 foodDurationMultiplier=0
 
+; If set to false, this option prevents food degrading over time - in other words, it retains its maximum benefit until it expires.
+foodDegradesOverTime=true
```
radmint, line 144: I think since a majority of our bool configurations default to false, we should continue that here to prevent user confusion and rename this to `disableFoodDegradation=false` or something similar.

ValheimPlus/GameClasses/Player.cs
```diff
@@ -135,77 +139,83 @@ private static void Postfix(ref Player __instance, ref Vector3 ___m_moveDir, ref
...
-            return false;
+            return il.AsEnumerable();
         }
 
-        private static float getModifiedDeltaTime(ref Player __instance, ref float dt)
+        private static float ComputeModifiedDT(float dt)
         {
-            float defaultDeltaTimeTarget = 1f;
-            float newDetalTimeTarget;
+            return dt / Helper.applyModifierValue(1.0f, Configuration.Current.Food.foodDurationMultiplier);
```
radmint, line 177: Is this retaining the same functionality of the original `getModifiedDeltaTime`?

radmint (changes requested):
Thank you! Just a couple of questions and a requested change.

ValheimPlus/GameClasses/Player.cs
cstamford: Yes, probably... subjectively, I'm not a fan of `yield return`. I think it makes the code harder to understand. But it does have a marginal performance benefit. The refactored code would look something like this (may not work!):

```cs
[HarmonyTranspiler]
public static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstruction> instructions)
{
    bool should_modify = Configuration.Current.Food.IsEnabled && Configuration.Current.Food.foodDegradesOverTime;

    foreach (CodeInstruction inst in instructions)
    {
        bool loads_health = inst.LoadsField(field_Food_m_health);
        bool loads_stamina = inst.LoadsField(field_Food_m_stamina);

        if (should_modify && (loads_health || loads_stamina))
        {
            inst.operand = field_Food_m_item;
            yield return inst;
            yield return new CodeInstruction(OpCodes.Ldfld, field_ItemData_m_shared);
            yield return new CodeInstruction(OpCodes.Ldfld, loads_health ? field_SharedData_m_food : field_SharedData_m_foodStamina);
        }
        else
        {
            yield return inst;
        }
    }
}
```

radmint: Okay yeah I think we'd rather have maintainability than a marginal performance increase  haha we can keep as is

ValheimPlus/GameClasses/Player.cs
cstamford: Yes! Same behaviour, slightly different approach. Given a setting of 100 (e.g. double duration), the refactored code accumulates time at half rate, while the old code accumulated time at the normal rate but changed the "refresh target" to be double. I thought this approach would be more scalable (since the target of 1 is hardcoded, future code that reads `m_foodUpdateTimer` may expect it to be within that range) but it wasn't the focus of the refactor and I can switch it back around if you'd like.

radmint: I think this is fine, just wanted to make sure we weren't changing any user-perceived functionality but I expect that @nxPublic will want to review this

radmint (approved):

nxPublic (approved):

PR #302, opened 2021-03-09
Added auto repair feature when interacting with workbench.

cstamford:
This contains the functionality from https://www.nexusmods.com/valheim/mods/51 (plus - code improved a bit to only spawn one sfx rather than a ton of them when repairing many items).

### DIFF
--- a/ValheimPlus/GameClasses/InventoryGUI.cs
+++ b/ValheimPlus/GameClasses/InventoryGUI.cs
@@ -1,5 +1,9 @@
 using HarmonyLib;
 using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Reflection.Emit;
 using UnityEngine;
 using UnityEngine.UI;
 using ValheimPlus.Configurations;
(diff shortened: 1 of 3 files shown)


ValheimPlus/GameClasses/InventoryGUI.cs
```diff
@@ -104,4 +108,77 @@ public static void Postfix(ref InventoryGui __instance)
...
+	public static class InventoryGui_UpdateRepair_Patch
+	{
+		/// <summary>
+		/// When we're in a state where the InventoryGui is open and we have items available to repair,
+		/// and we have an active crafting station, this patch is responsible for repairing all items
+		/// that can be repaired and then spawning one instance of the repair effect if at least one item
+		/// has been repaired.
+		/// </summary>
+		[HarmonyPrefix]
+		public static void Prefix(InventoryGui __instance)
```
radmint, line 161: I think this parameter requires `ref`

radmint (changes requested):
This looks great, just one suggested change

ValheimPlus/GameClasses/InventoryGUI.cs
cstamford: It would only require ref if `InventoryGui` was a value type. Since `InventoryGui` is a reference type, we can do things to its members without using ref. We would only need a ref if this was, for example, an argument and we wanted to change which object it pointed to, like this:

```cs
[HarmonyPrefix]
public static void Prefix(InventoryGui arg)
{
    arg = null;
}
```

This would do nothing.

```cs
[HarmonyPrefix]
public static void Prefix(ref InventoryGui arg)
{
    arg = null;
}
```

This would change the argument that is passed to the original function to be null.

radmint: so by not passing `ref` we'll still have the same desired affect when running `__instance.RepairOneItem()`?

cstamford: Yep, in this case, the behaviour would only be different if `InventoryGui` was a value type (in other words, a struct).

radmint (approved):

nxPublic (approved):

PR #303, opened 2021-03-09
Added a number of QoL changes.

cstamford:
* Death penalty can be tweaked/disabled
* Rested bonus per comfort level can be tweaked
* Radius for pieces to contribute to comfort can be tweaked

This brings in features from https://www.nexusmods.com/valheim/mods/50.

### DIFF
--- a/ValheimPlus/Configurations/Sections/BuildingConfiguration.cs
+++ b/ValheimPlus/Configurations/Sections/BuildingConfiguration.cs
@@ -5,7 +5,7 @@
         public bool noInvalidPlacementRestriction { get; set; } = false;
         public bool noWeatherDamage { get; set; } = false;
         public float maximumPlacementDistance { get; internal set; } = 5;
-        public float pieceComfortRadius { get; internal set; } = 0;
+        public float pieceComfortRadius { get; internal set; } = 10;
     }
 
 }
--- a/ValheimPlus/Configurations/Sections/BuildingConfiguration.cs
+++ b/ValheimPlus/Configurations/Sections/BuildingConfiguration.cs
@@ -5,7 +5,7 @@
         public bool noInvalidPlacementRestriction { get; set; } = false;
         public bool noWeatherDamage { get; set; } = false;
         public float maximumPlacementDistance { get; internal set; } = 5;
-
+        public float pieceComfortRadius { get; internal set; } = 0;
     }
 
 }
--- a/ValheimPlus/Configurations/Sections/PlayerConfiguration.cs
+++ b/ValheimPlus/Configurations/Sections/PlayerConfiguration.cs
@@ -8,7 +8,7 @@
         public bool disableCameraShake { get; internal set; } = false;
         public float baseUnarmedDamage { get; internal set; } = 0;
         public bool cropNotifier { get; internal set; } = false;
-        public float restSecondsPerComfortLevel { get; internal set; } = 0;
+        public float restSecondsPerComfortLevel { get; internal set; } = 60;
         public float deathPenaltyMultiplier { get; internal set; } = 0;
     }
 }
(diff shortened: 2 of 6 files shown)


cstamford:
Note that I did these in % rather than absolute values because that seemed to be the direction newer code was going - but happy to change it to be absolute instead.

ValheimPlus/Configurations/Sections/PlayerConfiguration.cs
```diff
@@ -8,6 +8,7 @@ public class PlayerConfiguration : ServerSyncConfig<PlayerConfiguration>
         public bool disableCameraShake { get; internal set; } = false;
         public float baseUnarmedDamage { get; internal set; } = 0;
         public bool cropNotifier { get; internal set; } = false;
-
+        public float restSecondsPerComfortLevel { get; internal set; } = 0;
```
radmint, line 11: I think this value would be better represented as its default float `= 60` and people can update it to what they like instead of a percentage modifier

ValheimPlus/Configurations/Sections/BuildingConfiguration.cs
```diff
@@ -5,7 +5,7 @@ public class BuildingConfiguration : ServerSyncConfig<BuildingConfiguration>
         public bool noInvalidPlacementRestriction { get; set; } = false;
         public bool noWeatherDamage { get; set; } = false;
         public float maximumPlacementDistance { get; internal set; } = 5;
-
+        public float pieceComfortRadius { get; internal set; } = 0;
```
radmint, line 8: I wonder if this would be better placed in the `[Items]` section. What do you think?

ValheimPlus/GameClasses/SE_Rested.cs
```diff
@@ -0,0 +1,63 @@
...
+        [HarmonyTranspiler]
+        public static IEnumerable<CodeInstruction> Transpiler(IEnumerable<CodeInstruction> instructions)
+        {
+            if (!Configuration.Current.Building.IsEnabled) return instructions;
+
+            List<CodeInstruction> il = instructions.ToList();
+
+            for (int i = 0; i < il.Count; ++i)
+            {
+                if (il[i].Calls(method_SE_Rested_GetNearbyPieces))
```
radmint, line 45: (for my transpiler knowledge) this is doing the same thing as `prefix return false` but that only skips the original isntead of original and other prefixes right?

ValheimPlus/Configurations/Sections/BuildingConfiguration.cs
radmint: Also I think this configuration would be better as the default radius of "10" and users can modify it how they like rather than a percentage

valheim_plus.cfg
```diff
@@ -343,6 +346,12 @@ baseUnarmedDamage=120
 ; When changed to true, you will not be permitted to place a crop within the grow radius of another crop
 cropNotifier=false
 
+; How many seconds each comfort level contributes to the rested bonus, in percentage.
+restSecondsPerComfortLevel=0
+
+; Change the death penalty in percentage - set to -100 to disable it.
```
radmint, line 352: For clarity: "set to -100 to disable the skill death penalty entirely" also would yo mind adding an example to how this is used? Like "setting the value to 10 would lower all skills by 10% on player death" or something.

radmint (changes requested):
Looks great! I think moving back to absolute values for some of these would be better but that's my only feedback!

valheim_plus.cfg
cstamford: It's actually reverse, so setting it to 50 would increase death penalties by 50%, and setting it to -50 would decrease death penalties by 50%... I can add examples or change how it works, what do you think?

radmint: No need to change functionality, examples would be great!

ValheimPlus/GameClasses/SE_Rested.cs
radmint: Awesome, thanks so much for taking the time to explain that!

radmint (dismissed):

nxPublic (approved):

PR #306, opened 2021-03-10
Properly apply radius increases of the CraftingStation and Ward to their EffectArea

drewaburden:
CraftingStation and Wards have a child GameObject that has an EffectArea and a collider. The mob spawning system uses this collision to check whether spawning is allowed, and the collision radius is not automatically updated with the CraftingStation/Ward radii. So, without this, mobs can still spawn inside those expanded areas. By adjusting the EffectArea as well, we can ensure that any modification to the ranges there apply to mob spawning as well.

### DIFF
--- a/ValheimPlus/GameClasses/CraftingStation.cs
+++ b/ValheimPlus/GameClasses/CraftingStation.cs
@@ -12,7 +12,7 @@ namespace ValheimPlus
         [HarmonyPatch(typeof(CraftingStation), "Start")]
         public static class WorkbenchRangeIncrease
         {
-            private static void Prefix(ref float ___m_rangeBuild, GameObject ___m_areaMarker)
+            private static void Prefix(CraftingStation __instance, ref float ___m_rangeBuild, GameObject ___m_areaMarker)
             {
                 if (Configuration.Current.Workbench.IsEnabled && Configuration.Current.Workbench.workbenchRange > 0) 
                 {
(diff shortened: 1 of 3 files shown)


nxPublic (approved):

nxPublic:
Also the next time please use ref on passed function arguments.

nxPublic:
My comment was not about the general C# behavior, this is about Harmony's behavior.

nxPublic:
If you do not pass the objects in the parameters as a ref, you cannot modify them. Its essentially a different instance then, permanently.

drewaburden:
You're saying Harmony explicitly creates a copy of the entire instance before passing it, if it's not marked with the ref keyword? Why would they do that? That seems really odd to me, and I can't think of any real benefit. Do you happen to have a link to documentation about that? I haven't been able to find anything.

cstamford:
That is only true for value types (e.g. primitive types and structs, see https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/value-types). In practice, the difference between `ref ReferenceType blah` and `ReferenceType blah` is that the former can have its assignment changed (e.g. `blah = null;`) and the caller will observe that change, while the latter would see no difference.

Harmony will do this (pseudocode) if you use `ref`:

```cs
float arg = original_arg;
foreach (PrefixHook hook in hooks)
{
    hook(ref arg);
}
```

And this (pseudocode) if you don't use ref:

```cs
float arg = original_arg
foreach (PrefixHook hook in hooks)
{
    hook(arg);
}
```

What this means changes depending on whether the arg is a reference or value type. In the case of a value type, modifying the value inside of `hook` will only change `arg` on the outside if it is marked as `ref`. For a reference type, because all reference types are pointers in C#, you don't receive a copy of the `arg` inside `hook`, you receive a copy of the pointer to the object. You would only need `ref` if you wanted to change the object that the reference type pointed to, like: `arg = null;` (and you wanted to propagate it to the next callers).

The `__instance` would only ever need a `ref` if it was a value type.

PR #307, opened 2021-03-10
Added ability to modify Smelter parameters similar to Blast Furnace and Kiln

drewaburden:
Base Smelter properties can now be modified like the Blast Furnace and Kiln properties.
Added new configuration for the base Smelter. It has the same properties as the Furnace, but it is under a new configuration section.

### DIFF
--- a/ValheimPlus/Configurations/Configuration.cs
+++ b/ValheimPlus/Configurations/Configuration.cs
@@ -15,6 +15,7 @@ namespace ValheimPlus.Configurations
         public FermenterConfiguration Fermenter { get; set; }
         public FireSourceConfiguration FireSource { get; set; }
         public FoodConfiguration Food { get; set; }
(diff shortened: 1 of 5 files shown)


nxPublic:
What are you supposedly trying to accomplish with this? Legit question.

drewaburden:
This allows modification of the tier 1 Smelter's properties. There was already the ability to modify the tier 2 Smelter (Blast Furnace) and the Kiln properties, but not the base Smelter. Since the Blast Furnace only smelts late-game ores, if you wanted to adjust the early/mid-game ore smelting, you couldn't without this.

radmint (approved):

nxPublic:
I have to apologize here. 

I have not been playing the game for a solid 3-4 weeks and at the time I wasn't even reaching the blast furnace. 
Until your post about this issue, I wasn't even aware about there being different smelters.

PR #308, opened 2021-03-10
Add ShieldConfiguration.cs to ValheimPlus.csproj

drewaburden:
For some reason, ShieldConfiguration.cs was not included in the project.


radmint:
this fix is already merged

PR #316, opened 2021-03-10
Merge items with existing stacks when picking up tombstone.

DavidKMartel:
This fixes a pretty big annoyance when you die. Currently Valheim tries to place items into the same inventory slot when  you pick up your tombstone. With this addition it will first try to merge with an existing stack then revert to the original behavior. So if you had one wood in your tombstone and you picked up one on the way, your new inventory would have a stack of two wood instead of two stacks of one wood.


nxPublic:
I have to make some more tests on this because it looks to me like it might remove some items on some conditions unintentionally.

ValheimPlus/GameClasses/Inventory.cs
```diff
@@ -80,4 +80,42 @@ public static void Prefix(string name, ref int w, ref int h)
...
+			if (Configuration.Current.Inventory.IsEnabled && Configuration.Current.Inventory.mergeWithExistingStacks)
+			{
+				List<ItemDrop.ItemData> list = new List<ItemDrop.ItemData>(fromInventory.GetAllItems());
+				foreach (ItemDrop.ItemData otherItem in list)
+				{
+					if (otherItem.m_shared.m_maxStackSize > 1)
+					{
+						foreach (ItemDrop.ItemData myItem in __instance.m_inventory)
+						{
+							if (myItem.m_shared.m_name == otherItem.m_shared.m_name && myItem.m_quality == otherItem.m_quality)
```
mle-ii, line 101: I would think most stackable items don't have a quality, but I suppose that doesn't account for modded items.  I haven't yet learned enough to mod valheim yet but are there other attributes we might need to check for items to ensure we don't stack things that have other attributes?  Though how do other mods deal with stackable items that have qualities and such?

ValheimPlus/GameClasses/Inventory.cs
```diff
@@ -80,4 +80,42 @@ public static void Prefix(string name, ref int w, ref int h)
...
+							{
+								int itemsToMove = Math.Min(myItem.m_shared.m_maxStackSize - myItem.m_stack, otherItem.m_stack);
+								myItem.m_stack += itemsToMove;
```
mle-ii, line 104: Should checks be done for weight?  I haven't played the game long enough but I imagine if we take a large stack of something heavy we'd have to go back into the inventory to drop that stack again to be able to run away from danger.

ValheimPlus/GameClasses/Inventory.cs
```diff
@@ -80,4 +80,42 @@ public static void Prefix(string name, ref int w, ref int h)
...
+								if (otherItem.m_stack == itemsToMove)
+								{
+									fromInventory.RemoveItem(otherItem);
+									break;
+								}
+								else
+								{
+									otherItem.m_stack -= itemsToMove;
```
mle-ii, line 112: Again, I've not written mods, but does this move all patch also run the default move all after this is run?  I ask as this will likely leave items in the dropped inventory that I'd still want to try to pick up.

mle-ii: Ah, I see from your original comment that it sounds like it does your code and then does the default code.

ValheimPlus/GameClasses/Inventory.cs
mle-ii: Though I suppose the default might not check this weight, but then this makes it much more of a chance of picking up too much weight.

nxPublic (dismissed):

nxPublic (approved):

ValheimPlus/GameClasses/Inventory.cs
DavidKMartel: Yes it doesn't make sense for stackable items to have a quality but the original decompiled code also has this check so it should stay.

I think in the past when minecraft had a limited number of item ids the damage field was abused to cram many items into the same item id. As long as the original code supports this abuse we should support it too.

ValheimPlus/GameClasses/Inventory.cs
DavidKMartel: The game currently doesn't check for weight. I've been thinking about this but I think we should add something smarter than just stop when you hit the weight limit. If we do add a check to stop at the weight capacity I would like it to prioritize important items such as weapons, armor and food. It would suck to only pick up your stack of ore and be left defenseless.

ValheimPlus/GameClasses/Inventory.cs
DavidKMartel: https://harmony.pardeike.net/articles/patching-prefix.html#changing-the-result-and-skipping-the-original

A prefix patch with a void return type will always pass on to other patches and then the original method. You can return false to bypass the original method.

PR #338, opened 2021-03-13
Fix for unix servers with spacing in folder paths

havedill:
Without quoting this unix servers wont start the server if their install dir contains spaces.


nxPublic:
We had this in the server start before and it cause issues for a number of our alpha testers.

havedill:
@nxPublic , i added a commit which attempts to export the path the regular way, but if it fails it attempts to use quotations. Working on my server with spacing in the path

nxPublic (approved):

PR #373, opened 2021-03-14
MegingjordBuff fix

adrian-wong:
Currently, se.ModifyMaxCarryWeight() can be called more than

With this quick fix, even if "beltstrength" is not an unique statusEffect, the weight should only be modified once.

-Mist


nxPublic:
Good catch!

nxPublic (approved):

PR #378, opened 2021-03-14
Updating start_server_bepinex.sh for 0.9.5.5 fix

Zxurian:
the changes to `start_server_bepinex.sh` regarding `LD_LIBRARY_PATH` were causing V+ not to load, thus resulting in the server reporting an incompatible version error.

This pull request contains the following changes
*Change 1*
Changed
```
LD_LIBRARY_PATH=`./linux64:$LD_LIBRARY_PATH` && export LD_LIBRARY_PATH || export LD_LIBRARY_PATH=./linux64:"$LD_LIBRARY_PATH"
```
to
```
export LD_LIBRARY_PATH="${VALHEIM_PLUS_PATH}/linux64":${LD_LIBRARY_PATH}
```
I couldn't see any reason for the first statement other than trying two different ways to export the same variable, so reverted it back to the pre `0.9.5.5` way.

*Change 2*
The method of using `${PWD}` breaks when calling the script from a directory other than the one `start_server_bepinex.sh` is in. This is a problem when run through third party platforms like LGSM. The new `${VALHEIM_PLUS_PATH}` reference will make sure that the path is absolute relative to the location of `start_server_bepinex.sh` regardless of where it's called from, ensuring compatibility with third party launchers.

*Change 3*
the file contained uses of both backticks and `$()` notation. Standardized all references to `$()` notation to be POSIX compliant.

### DIFF
--- a/resources/unix/start_server_bepinex.sh
+++ b/resources/unix/start_server_bepinex.sh
@@ -24,14 +24,18 @@ server_public=1
 
 # The rest is automatically handled by BepInEx for Valheim+
 
+# Set base path of start_server_bepinex.sh location
+export VALHEIM_PLUS_SCRIPT=$(readlink -f "$0")
+export VALHEIM_PLUS_PATH=$(dirname "$VALHEIM_PLUS_SCRIPT")
+
 # Whether or not to enable Doorstop. Valid values: TRUE or FALSE
 export DOORSTOP_ENABLE=TRUE
 
 # What .NET assembly to execute. Valid value is a path to a .NET DLL that mono can execute.
-export DOORSTOP_INVOKE_DLL_PATH="${PWD}/BepInEx/core/BepInEx.Preloader.dll"
+export DOORSTOP_INVOKE_DLL_PATH="${VALHEIM_PLUS_PATH}/BepInEx/core/BepInEx.Preloader.dll"
 
 # Which folder should be put in front of the Unity dll loading path
-export DOORSTOP_CORLIB_OVERRIDE_PATH=./unstripped_corlib
+export DOORSTOP_CORLIB_OVERRIDE_PATH="${VALHEIM_PLUS_PATH}/unstripped_corlib"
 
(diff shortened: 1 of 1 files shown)


havedill:
I had a fix written up in https://github.com/valheimPlus/ValheimPlus/pull/371

Zxurian:
> ... The reason it tries two different ways to export, is if teh unix path contains spaces, the pre .9.5.5 way breaks and wont launch the server

instead of trying it two different ways, isn't there a way to reliably handle it regardless of spaces? not at home so can't check my solution against spaces in path until then, but seems like *nix should have a way to safely escape paths without requiring an if. I'm certainly no bash/shell expert, but i'll take a look and update or adjust mine to match yours if there is no way.

havedill:
Originally i had merged a request which simply put the variable in quotations, but nx said that it was breaking things for some people who didn't have spacing in the path. That's why i opted to try both methods

thelostgrimm:
i did the change and the steam.so fails to start. very new to linux and im glad i found this

nxPublic:
@havedill , @Zxurian Would be nice if you guys came to a final conclusion on this issue since I personally don't have a Unix environment to test these solutions.

havedill:
What do you think a @Zxurian ? Honestly just setting the export to `
  export LD_LIBRARY_PATH=./linux64:"$LD_LIBRARY_PATH" ` in my opinion should have worked for both cases. But since it works for me i'm not sure what issues it was causing for other users

Zxurian:
@havedill was missing double quotes around a few of the path references that was breaking with spaces in mine. I just pushed an update that double quotes any reference to paths, and tested it on my server against server files being in a path with / without spaces, ran fine both times.
@nxPublic updated commit Tested with/without spaces. I haven't looked at the companion `start_game_bepinex.sh` though as I play on a Win10 machine, only host server on Ubuntu.

havedill:
@nxPublic  i've also tested the above patch with and without spacing in my structure and see no issues now on ubuntu 20.04

nxPublic (approved):

PR #392, opened 2021-03-17
Add Auto Fuel feature

healiha:
Allow the Kiln, Smelter, Furnace, Torches and fire places to pull fuel (wood, coal, ores,...) from nearby chests.

It uses the same range as the autoDrop feature.

Due to the way the game is made, we can't add a specific amount for fuel.
It means each item used as fuel we send something over the network, which isn't the best due to the limited dataRate.

To avoid any unecessary trouble like going over the fuel limit, only the peer having ownership of the current zone will have this feature enabled.

_Note: `autoDepositRange` has been renamed to `autoRange` for clarity_

### DIFF
--- a/ValheimPlus/GameClasses/Fireplace.cs
+++ b/ValheimPlus/GameClasses/Fireplace.cs
@@ -1,5 +1,9 @@
 ﻿using HarmonyLib;
+using System;
 using System.Linq;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Reflection.Emit;
 using ValheimPlus.Configurations;
 
 namespace ValheimPlus.GameClasses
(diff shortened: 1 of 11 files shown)


healiha:
Added auto-fuel for Firesources 😃

ValheimPlus/GameClasses/Fireplace.cs
```diff
@@ -40,20 +44,67 @@ private static void Postfix(ref double __result, ref Fireplace __instance)
...
+                if (!Configuration.Current.FireSource.IsEnabled || !Configuration.Current.FireSource.autoFuel) return instructions;
+
+                List<CodeInstruction> il = instructions.ToList();
+
+                for (int i = 0; i < il.Count; i++)
+                {
+                    if (il[i].Calls(method_Fireplace_UpdateState))
+                    {
+                        il.Insert(++i, new CodeInstruction(OpCodes.Ldarg_0));
+                        il.Insert(++i, new CodeInstruction(OpCodes.Call, method_addFuelFromNearbyChests));
```
Belphemur, line 79: Not inside IsOwner check

Should be after instance `bool ZNetView::IsOwner()` before the code check for the current value of the fuel.

healiha: Corrected!

_PS: It stings knowing that I was the one who told you that on discord 😂_

Belphemur: We did come up with the fix together though 😛

nxPublic (approved):

nxPublic:
Outstanding work! ❤️

PR #402, opened 2021-03-19
Level with stamina usage

ahnguyen09:
It would nice to have the current level show with the experience gain notification. This helps you check your current skill level quickly.
image


nxPublic (changes requested):
Please resubmit your PR without the project file changes and dependency changes.

It would also be nice to make it a config option.

I like this a lot though!

ahnguyen09:
@nxPublic I did not make it clear but this is an enhancement to the HUD "experienceGainedNotifications" option. Do you still think that it is worth adding another option to display level with the gains? Or we can just include it when experienceGainedNotifications is set to true.

nxPublic:
Yeah you are probably right. Just changing it is a improvement to giving a option to turn it off doesn't make that much sense.

ahnguyen09:
Will resubmit without project changes just code change to skill.cs

PR #408, opened 2021-03-20
Added projectile modification section, applies to bow and javelin

ahnguyen09:
This mod allows users to change how fast and accurate projectiles are. This mod started with trying to enhance bow gameplay.

I could not remove the delay on mouse release, but this is the next best thing. Your released arrows can be like a laser beam without affecting the damage.

I tested with +200% projectileVelMaxCharge and it felt pretty good when you also decrease the variance (increase accuracy)


ahnguyen09:
Upon further testing, the current code changes all projectiles. So Greydrawfs rocks become bullets, which is not great 😄 . I will have to change to code to a little.

healiha:
@ahnguyen09 You should try changing the `Prefix` condition to `if (Configuration.Current.ProjectileFired.IsEnabled && __instance.m_character.IsPlayer())`, it should give the expected result.

ahnguyen09:
@healiha thank you for the feedback. I figured if the users want to tweak the Greydrawf projectile, they should also be able to. The Greydrawfs could use some help 😄 . The code a bit longer but not a big deal.

healiha:
@ahnguyen09 Looking at the code, you would have to define a list of names you want it to be applied since this function is called for all `Humanoid`. Greydwarfs are one thing but Fulings spears being bullet-like... 😅

ahnguyen09:
@healiha Yeah you are right also rock throw troll, many star Drager archer, and so on... 😨

ValheimPlus/GameClasses/Attack.cs
```diff
@@ -59,4 +59,48 @@ private static void Postfix(ref Attack __instance, ref float __result)
...
+    public static class Attack_ProjectileAttackTriggered
+    {
+        private const float DEFAULT_PROJECTILE_VEL_MIN_CHARGE = 2f;
+        private const float DEFAULT_PROJECTILE_VEL_MAX_CHARGE = 50f;
+
+        private const float DEFAULT_PROJECTILE_VARIANCE_MIN_CHARGE = 20f;
+        private const float DEFAULT_PROJECTILE_VARIANCE_MAX_CHARGE = 1f;
+
+        private const float MAX_VALUE = 1e+6f;
+
```
nxPublic, line 76: Remove all consts and use the configuration instead to determine defaults

ValheimPlus/Configurations/Sections/ProjectileFiredConfiguration.cs
```diff
@@ -0,0 +1,17 @@
...
+        public float playerProjectileVelMaxCharge { get; internal set; } = 0;
+
+        public float playerProjectileVarMinCharge { get; internal set; } = 0;
+        public float playerProjectileVarMaxCharge { get; internal set; } = 0;
+
+        public float projectileVelMinCharge { get; internal set; } = 0;
+        public float projectileVelMaxCharge { get; internal set; } = 0;
+
+        public float projectileVarMinCharge { get; internal set; } = 0;
+        public float projectileVarMaxCharge { get; internal set; } = 0;
```
nxPublic, line 15: The configuration should contain all default values.

valheim_plus.cfg
```diff
@@ -692,3 +692,34 @@ alignToggle=F7
 
 ; Key to change the default alignment
 changeDefaultAlignment=F6
+
+[ProjectileFired]
```
nxPublic, line 696: These values should be set to the default values.

nxPublic (changes requested):

nxPublic:
Seems like a good thing to have to improve bow gameplay!

healiha:
I like the idea but it might be even better if we could add an option to change velocity and accuracy bease on current bow level to see a linear improvement.
What do you think?

ahnguyen09:
@healiha I like that the idea of ramping max velocity with level. It gives you a minor reward as you level up. I look into adding it.

ahnguyen09:
List of updates:

1.  Change mod class name to end with _Patch to conform with existing style
2. Change the namespace of the attack.cs file to conform with existing style
3. removed the constants in the mod class and moved it into configurations
    a. The config is now changes value instead of change percentage increase/decrease
    b. I like this better anyway now that scaling is added.
4. added ability to enable linear scaling of velocity and accuracy based on level
    a. User can now have a lot of customization around projectile scaling (or not)

ValheimPlus/GameClasses/Attack.cs
```diff
@@ -59,4 +60,57 @@ private static void Postfix(ref Attack __instance, ref float __result)
...
+                else
+                {
+                    __instance.m_projectileVelMin = Configuration.Current.ProjectileFired.projectileVelMinCharge;
+                    __instance.m_projectileVel = Configuration.Current.ProjectileFired.projectileVelMaxCharge;
+
+                    __instance.m_projectileAccuracyMin = Configuration.Current.ProjectileFired.projectileVarMinCharge;
+                    __instance.m_projectileAccuracy = Configuration.Current.ProjectileFired.projectileVarMaxCharge;
+                }
+
+                float maxValue = 1e+6f;
```
healiha, line 106: IMO, since this is a constant you should move it to the class scope.

healiha (changes requested):
It looks promising 😃 

Remember to mark requested changes as resolved once done ;)

ahnguyen09:
> It looks promising 😃
> 
> Remember to mark requested changes as resolved once done ;)

I think i just resolved it. I a still a little new to Contributing.

ValheimPlus/GameClasses/Attack.cs
ahnguyen09: Yeah I probably when overboard with the change. @nxPublic told me to remove the constant and make them config so I thought it is bad practice to have class const. I rather have it as class const as well.

nxPublic:
Please resolve the merge conflicts and add your configuration file changes to the [vplusconfig.json](https://github.com/valheimPlus/ValheimPlus/blob/development/vplusconfig.json).

Everything else seems merge ready to me :)

ahnguyen09 (commented):
was done

nxPublic:
I think we need a lot better descriptions on these different configurations, we should also consider to split up the project changes into 2 different main categories, monster projectiles and player projectiles. Hit me up on discord to discuss this if you like :)

ahnguyen09 (commented):
It looks good to me.

nxPublic (approved):

PR #409, opened 2021-03-20
Added: PVP Switch and Skip Intro

J-Pster:
This Pull Request does the addition of two things.

# PVP Switch
PVP Switch is a function that sets PVP to either on or off for all players on the server.
It also prevents them from switching between PVP On and PVP Off mode when it is enabled.

The configuration is simple:
```
; This is the area that changes the PVP mode, set "pvpveactivator" to false to disable it.
; Change this to true to enable the PVP Switch section, or to false to disable the PVP Switch section.
pvpveactv=false
; [PVP Switch] Change this to true for PVP enabled and to false for PVP disabled.
pvpveswitch=false
```

# Skip Intro
This addition causes the introduction of a new character, that "bird" that brings new characters into the world to be removed when activated.

```
; Change this to true to disable the initial animation of a new player, that "Bird" you bring down to earth.
skipintro=false
```

### DIFF
--- a/ValheimPlus/Configurations/Sections/PlayerConfiguration.cs
+++ b/ValheimPlus/Configurations/Sections/PlayerConfiguration.cs
@@ -14,5 +14,8 @@
         public float guardianBuffDuration { get; internal set; } = 300;
         public float guardianBuffCooldown { get; internal set; } = 1200;
         public bool autoEquipShield { get; internal set; } = false;
+        public bool pvpveswitch { get; internal set; } = false;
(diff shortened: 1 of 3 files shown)


nxPublic:
We are currently working on a extra pvp addition to Valheim Plus.

More on this later today with the update to the repository.

CastBlacKing:
> 
> 
> We are currently working on a extra pvp addition to Valheim Plus.
> 
> More on within the next few days with the update to the repository.

:/

PR #420, opened 2021-03-22
Fermenter, Windmill and Spinning Wheel customization

healiha:
Added Fermenter, Windmill and Spinning Wheel customization. Those are the last two "smelters" we were missing customization for.

Also removed nested `if` to reduce time complexity.


nxPublic (approved):

nxPublic:
Outstanding work, like every time so far. ❤️

Issue #205, opened 2021-03-02
[Feature Request] Mute sound when in background


Xiaminou:
Add an option for the client to mute Valheim's sound when it's in the background.
It should be in the base game but since it's not I'm hoping you guys can help with that.

radmint:
![](https://github.trello.services/images/mini-trello-icon.png) [Mute sound when in background](https://trello.com/c/lTa26r9D/39-mute-sound-when-in-background)

nxPublic:
This has been added for 0.9.7 by #430

PR #430, opened 2021-03-23
Mute audio when application loses focus

baconparticles:
Satisfies Feature Request #205
May want to augment this to be client side only or a ui-setttings toggle instead of a config file option.

### DIFF
--- a/ValheimPlus/GameClasses/EventSystem.cs
+++ b/ValheimPlus/GameClasses/EventSystem.cs
@@ -1,0 +1,24 @@
+﻿using HarmonyLib;
+using UnityEngine;
+using UnityEngine.UI;
+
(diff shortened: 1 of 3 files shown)


healiha:
Thank you for making this pull request!

As you said, we will be looking into adding this to the game settings UI as in my opinion it does not really have it's place in the V+ config file.

baconparticles:
Sounds good. I am looking into how to do that now, but I am new to .NET and dnSpy so it's taking some time.

baconparticles:
Removed config based solution in favor of settings UI based solution.
Known issues: EventSystem does not push OnApplicationFocus while in the settings dialog itself

nxPublic:
Nice work! ❤️

nxPublic (approved):

PR #433, opened 2021-03-23
Version Messages

CastBlacKing:
**This implementation adds more variation to the version notification.**

### There are now 3 different messages, new, same and old.

* When the version published in the repository is the same as the one being run the ``same`` message is sent.
* When the version published in the repository is newer than the one you are running the ``new`` message will be sent.
* When the version published in the repository is older than the one you are running the ``old`` message is sent

Read what the ``same, new and old`` messages represent by looking at the modified files in this PR.


nxPublic:
I don't think a change like this would make sense.

We only supply newer versions to alpha testers, the version check should only fail if there is no internet connection and that is handled already.

CastBlacKing:
> 
> 
> I don't think a change like this would make sense.
> 
> We only supply newer versions to alpha testers, the version check should only fail if there is no internet connection and that is handled already.

Okay.

PR #446, opened 2021-03-24
Allow manual switching to any shield with autoEquipShield active

exscape:
Previously, the code checks whether you have a one-handed weapon and equips the best shield available, even if the user is manually trying to equip a different shield.  
(May also fix issues when having two identical shields with different durability, rather than two different shield types.)


nxPublic (approved):
Good catch! Thank you 👍

PR #458, opened 2021-03-25
Fix item disposal

healiha:
Changed the method used to set the time before an item should be destroyed when lying on the ground.

The previous method wasn't fail proof as it would throw errors on newly created servers due to the way the instance time is computed.


Mixone-FinallyHere (approved):
lgtm

nxPublic (approved):

PR #466, opened 2021-03-25
Allow adjustments of fall damage

exscape:
Allows both for scaling fall damage via a percentage, and for changing the maximum (default is 100).


exscape:
Updated to remove the suggestion from todo.json.

nxPublic (approved):

PR #471, opened 2021-03-26
Update FermenterConfiguration.cs

rr222cy:
The default fermenterDuration was set to 10, changed this back to default. Also changed autoDeposit and the likes to false, since these are settings not even available in the vanilla experience I suggest that the default value for V+ should be false, to avoid confusion for new users and so that they have to opt-in, instead of the opposite.


healiha:
I overlooked that part when creating the PR, good thing you noticed it!

healiha (approved):

nxPublic (approved):

PR #473, opened 2021-03-26
Update valheim_plus.cfg

rr222cy:
According to the Wiki, the production speed for 1 piece of coal is 16 sec: https://valheim.fandom.com/wiki/Charcoal_kiln

### DIFF
--- a/valheim_plus.cfg
+++ b/valheim_plus.cfg
@@ -428,7 +428,7 @@ dontProcessFineWood=false
 dontProcessRoundLog=false
 
 ; The time it takes for the Kiln to produce a single piece of coal in seconds.
-productionSpeed=30
+productionSpeed=16
 
 ; Instead of dropping the items, they will be placed inside nearby chests.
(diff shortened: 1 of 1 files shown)


healiha:
I just checked that, the wiki doesn't have the right value.

According to the game default value, it should be 15.

valheim_plus.cfg
```diff
@@ -428,7 +428,7 @@ dontProcessFineWood=false
 dontProcessRoundLog=false
 
 ; The time it takes for the Kiln to produce a single piece of coal in seconds.
-productionSpeed=30
+productionSpeed=16
```
healiha, line 431: Should be 15 as per the default game settings

healiha (changes requested):

valheim_plus.cfg
rr222cy: Thanks for checking, you're right!

nxPublic (approved):

PR #474, opened 2021-03-26
Update KilnConfiguration.cs

rr222cy:
According to the wiki, the default production speed for 1 piece of coal is 16 sec: https://valheim.fandom.com/wiki/Charcoal_kiln

### DIFF
--- a/ValheimPlus/Configurations/Sections/KilnConfiguration.cs
+++ b/ValheimPlus/Configurations/Sections/KilnConfiguration.cs
@@ -2,7 +2,7 @@
 {
     public class KilnConfiguration : ServerSyncConfig<KilnConfiguration>
     {
-        public float productionSpeed { get; internal set; } = 10;
+        public float productionSpeed { get; internal set; } = 16;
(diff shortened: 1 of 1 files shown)


ValheimPlus/Configurations/Sections/KilnConfiguration.cs
```diff
@@ -2,7 +2,7 @@
 {
     public class KilnConfiguration : ServerSyncConfig<KilnConfiguration>
     {
-        public float productionSpeed { get; internal set; } = 10;
+        public float productionSpeed { get; internal set; } = 16;
```
healiha, line 5: Should be 15 as per the default game settings

healiha (changes requested):

ValheimPlus/Configurations/Sections/KilnConfiguration.cs
rr222cy: Thanks for checking, you're right!

nxPublic (approved):

Issue #365, opened 2021-03-14
[FEATURE REQUEST] Time of day


Lizzardd:
Would be great to have a time of day clock displayed on the hud.
This one exists, but either doesn't work or isn't compatible with V+
https://www.nexusmods.com/valheim/mods/85

ShammyLevva:
That Clock mod works fine on my install, maybe you didn't install the pre-requisites? Agreed though it would be nice to see that rolled into Plus.

VikingStocky:
I agree. V+ has everything I need, except an in game clock.

I'd prefer not to download other Mods at this time.

I'll wait for V+ to add it. I'm sure they will eventually.

nxPublic:
This suggestion has been added to http://valheim.plus/todo under Hud as "In-game Clock".

PR #489, opened 2021-03-29
Added In-game clock

ahnguyen09:
image
#365 There was a request for In-Game clock by a lot of people (number 1 on the HUD according to the https://valheim.plus/todo).

Well here it is. 

- Allow both 24hr (default) time and AM-PM format.
- Allow user to change Font Size
- Allow user to customize r,g,b,a channels


ahnguyen09:
TODO: Need to test with different resolution and look CanvasScaler canvasScalerComponent.

ahnguyen09:
@nxPublic It seems like by default Text has automatic UI scaling. I changed the resolution and there doesn't seem to be any problems. I also tried in windows mode and the text scales to be smaller or bigger according to the window.

nxPublic (approved):

Issue #222, opened 2021-03-04
[Feature Request] Berries yeld increase option


shilliash:
As subjected it is kind of self explanatory. I am considering weird player receives only 1 berry from entire bush - would be awesome if we have a chance to modify that to 2 or 3 pcs from one bush via V+

radmint:
![](https://github.trello.services/images/mini-trello-icon.png) [modifiers for production of berries on all kinds of bushes](https://trello.com/c/t56Nebci/51-modifiers-for-production-of-berries-on-all-kinds-of-bushes)

nxPublic:
This suggestion has been added to http://valheim.plus/todo under Items as "Modifiers for Berries and Bushes".

PR #493, opened 2021-03-30
added pickable item and add ref tutorial

ahnguyen09:
#222 Add ability to modify the amount drop of some pickable items.


ahnguyen09:
A better implementation exist #494.

PR #494, opened 2021-03-30
Allow modifiying Pickable yield

exscape:
Pickables are items that appear when you press the Use button on something, e.g. a berry bush, a branch or a carrot.
This patch allows you to modify the yield of most pickables.

Probably needs a bit of extra alpha testing, but everything I've tested has worked both locally and on a dedicated server.


ahnguyen09:
@exscape Looks good man. Much better than what I did. I will let you take over and I will remove my PR #493.

exscape:
> @exscape Looks good man. Much better than what I did. I will let you take over and I will remove my PR #493.

Thanks, but wow, that's awful timing to have us submit them as #493 and #494 after it's been a request for a few weeks! :)

nxPublic (approved):

Issue #414, opened 2021-03-21
[FEATURE REQUEST] Disable rain/weather


joklbash:
**Is your feature request related to a problem? If so, please describe.**
Weather in this game can be a real downer. It's dark enough at times that venturing out is a pain. When it rains, I generally log off for a while rather than deal with it. It's cool the first couple times you see it and even when you're out to sea. After 100+ hours of play, it is just a pain to deal with when you're trying to build or just explore around.

**Describe the solution you'd like**
I'd like an option on dedicated server to disable rain or maybe reduce how often it can occur. For me, I'd like to just disable rain and maybe even fog altogether. Others might rather have something like rain not occurring more often than every 7 days or something.

**Additional context**
Single player appears to support changing the weather.

DogBeatr:
This would be the best quality of life addition.  The rain storms for us are fine, it's the fog that is really annoying.  It lasts WAAAAAY too long and happens way too often.

nxPublic:
Messing around in the EnvMan class is a really bad idea, specially for a mod used as much as ours.

We simply cannot fiddle around in that system, its to complex and results in to many issues for other mods.

PR #501, opened 2021-03-31
added mod for disable fog and brightness adjustment

ahnguyen09:
I didn't know a good way to separate the two PR and can be added at the same time, so it will be lumped into one.

1. Add the option to remove fog: #414 
2. It was unplayable sometimes at night (and the sun is out IRL). So here is the option to modify brightness. You can also think of it as a way to change gamma.
  
Brightness was handle by converting all of the base game color to HSV and changing the V (brightness).


ahnguyen09:
Update to code:

- Removed morning, day and evening options because it affected brightness in a weird way. Will need a lot more development it seems, so let just change the night brightness for now.
- Change the scaling to make the number feel more intuitive (+100 ~ 2x in brightness)

nxPublic (approved):

PR #502, opened 2021-04-02
Add auto-split functionality on alt+click

exscape:
Alt+click on a stack in a container will take as much of the stack as
you can carry without becoming encumbered.
If you already are encumbered, alt+click on a stack in your inventory
will move as *little* as possible to become unencumbered.

If you're in your inventory with no container open, alt+click will
instead *drop* as little as possible to become unencumbered.

Holding shift as well will toggle how the move happens.
The default is to "auto-move", which works essentially like ctrl+click
in the base game: it moves to partial stacks first, then to an empty slot
if still needed.

Holding alt+shift will split like the game's split feature and make the
user place the item(s) where they wish.

The code is fairly well tested.


exscape:
Added the right-click-to-place feature as discussed with @nxPublic.  

Also removed the original suggestion from todo.json, though only half is implemented, not sure if it should remain:

> Right clicking to take a single item from a stack (or place from a held stack into an empty slot) at a time is something I find myself wanting to do very often, and in line with functionality from Valheim predecessors such as Terraria

Since right-clicking a stack uses the item, implementing this without e.g. a modifier would make it impossible to eat food or equip items without having them in the numbered slots.  
This patch removes right-click-to-eat only if you're currently dragging an item (and the config option is set), and leaves it be otherwise.

exscape:
Updated to fix a minor issue when using this feature to move equipped items.

Ironically I forgot to mention that I haven't tested this fully yet, but I'd say 90% or so. Maybe a bit more now.

exscape:
Updated again to roll back the right-click-to-place feature, unfortunately.

I started working on shift+right-click to pick items up, but doing so has issues. Most notably, it's impossible to pick up from multiple different stacks without *major* changes to the game's inventory code, as it keeps track of a "from" stack during drags (and this counts as dragging), and any actual changes to the stack only happens after we place an item down. So we can't move from multiple stacks unless we track everything ourselves and basically rewrite the item move/drag system.

nxPublic:
We just came to the conclusion that the possible gain of this feature is not worth risking further incompatibility with other mods to this severe degree.

PR #505, opened 2021-04-03
Added issues.json for website's upcoming issues page - needs maintaining manually.

ncjsvr:
We could pull this from Github's API but then every open "bug" get's fed in to the page and I don't agree with showing bugs that can't be replicated or verified.


EldRoos:
Add lable confimed and filter by that then.

PR #511, opened 2021-04-05
Fixed logic error in disableCookingStation when crafting from chest

Dimitro:
Option was working backwards.


nxPublic:
Good catch!

PR #514, opened 2021-04-06
Fix furnace->allowAllOres option for iron smelting

Dimitro:
Iron Ore is present in the game files but it's unobtainable, game using Scrap Iron instead.


nxPublic:
Another good catch! 👍

nxPublic (approved):

AidanGG:
Should this be semi-reverted now that Iron Ore and Copper Scrap are in the game?

PR #522, opened 2021-04-08
Protected Pets - Patch

Menthus123:
Fixed a couple of issues, updated config and removed Protected Pets suggestion from todo.json.


Mixone-FinallyHere (approved):
Looks great! :D

nxPublic (approved):

PR #551, opened 2021-04-21
Added option to make shared pins be enforced, i.e. NO private pins

Mixone-FinallyHere:
Works retroactively for existing pins without modifying data structure


Mixone-FinallyHere:
Tracked in #553

PR #553, opened 2021-04-23
Config UI

Mixone-FinallyHere:


Mixone-FinallyHere:
Made in game mod settings get applied, in theory done. 

Needs testing and beautifying before release, but yeah, essentially done

PR #564, opened 2021-05-01
Added support for Strings in Config File & Added the possibility to change the OnSpawn Text Message

MrTruth0:


nxPublic:
Thank you for the effort and PR, its good clean code. However we don't think this to be a quality of life addition consistent with what we want to achieve with the Valheim Plus project. Your changes don't contain any noteworthy gameplay related changes.

Feel free to get in touch with us on [discord](http://discord.gg/valheimPlus).

PR #578, opened 2021-05-19
Added Disable Encumbered option

MrTruth0:


nxPublic:
I think you should change this into a Postfix with the same functionality, otherwise other mods that are installed would potentially not be able to use this method when its enabled at all.

MrTruth0:
The Patch overwrite the IsEncumbered return value, with a postfix will throw an Exception

Mixone-FinallyHere:
> The Patch overwrite the IsEncumbered return value, with a postfix will throw an Exception

How so? 

```cs
[HarmonyPatch(typeof(Player), nameof(Player.IsEncumbered))]
public static class Player_DisableEncumbered_Patch
{
    private static void Postfix(ref bool __result)
    {
        if (Configuration.Current.Player.IsEnabled && Configuration.Current.Player.disableEncumbered)
        {
            __result = false;
        }
    }
}
```

That should work without erros no?

MrTruth0:
> > The Patch overwrite the IsEncumbered return value, with a postfix will throw an Exception
> 
> How so?
> 
> ```cs
> [HarmonyPatch(typeof(Player), nameof(Player.IsEncumbered))]
> public static class Player_DisableEncumbered_Patch
> {
>     private static void Postfix(ref bool __result)
>     {
>         if (Configuration.Current.Player.IsEnabled && Configuration.Current.Player.disableEncumbered)
>         {
>             __result = false;
>         }
>     }
> }
> ```
> 
> That should work without erros no?

Ops ! Yeah it works

PR #579, opened 2021-05-19
Update valheim_plus.cfg setting fix, and added missing setting to vplusconfig.json

stolenvw:
Fixed Kiln ignore Private Area Check setting so it reads it and dont use default setting if its been changed
Added missing shareAllPins setting to vplusconfig.json


Mixone-FinallyHere (approved):
Awesome looks great

PR #595, opened 2021-05-25
Global Pins Synchronization

f1ames0ff:
### Global Pins Synchronization

Feature to sync navigational pins _(circle, fire, etc.)_ from all players, connected to a server between each other.
All pins will be merged on the server and after send to all clients.
On a client minimap will be updated accordingly with the merged pins from the server.

I'm making draft PR because the feature is working correctly, but I have **couple of questions:**

1. What hook I need to use to run my core only one time after client map fully initialised?
2. Config doesn't update properly, doesn't add my new property and I don't understand why.


Mixone-FinallyHere (commented):
@nxPublic  I am checking with f1am3d please dont merge yet

ValheimPlus/RPC/VPlusMapGlobalPinSync.cs
```diff
@@ -0,0 +1,409 @@
...
+            {
+                ZLog.LogWarning("MapPinsSync: Got empty map pin package from server.");
+                return;
+            }
+            
+            var pos = mapPinPkg.ReadVector3();
+            var radius = mapPinPkg.ReadSingle();
+            var pin = Minimap.instance.GetClosestPin(pos, radius);
+
+            CachedPin cachedPin = ConvertRawToCachedPin(pin);
```
sbtoonz, line 105: This always returns null

sbtoonz (changes requested):

nxPublic:
Since this is now months old, its fair to assume no progress has been made.

f1ames0ff:
> Since this is now months old, its fair to assume no progress has been made.

Progress was made and it's working. But no one is checking PRs here

nxPublic:
Please get in touch with me on discord then, if I remember correctly you mostly talked with Mixone who is currently on holidays.

Issue #455, opened 2021-03-25
[BUG] Translating objects in advanced building mode yields inconsistent results


dewiniaid:
- **OS:** Windows Client
- **Valheim Game Version:** 0.148.6
- **V+ Mod Version:** 0.9.6

**Describe the bug**
When using advanced building, translating an object back and forth between two different locations does not always return it to its starting location.  It seems like a given input to move the object is processed more than once.

My entire reason for experimenting with Building Extensions is to be better able to snap objects together (particularly stone walls with an origin below ground), so this makes doing so consistently particularly difficult.

**To Reproduce**

Steps for the devs to reproduce the behavior:
1. Hit F1 while building to toggle advanced building.  This is best done if the object is currently snapped to another object to better see where it started.
2. Do Left (or Shift+Left) 10 times
3. Do Right (or Shift+Right) 10 times.
4. Observe that the object is no longer in its original position.
5. Retry a few times if unable to reproduce on the first attempt.

**Expected behavior**
Movement actions in advanced building should 1:1 cancel out.

**Troubleshooting**
Not really applicable here.

**Additional context**
No other mods other than required dependencies; installed through Vortex (aka Nexus Mod Manager).

Abrackadabra:
+1, I was surprised when I discovered this.

Another issue seems to be that increaseScrollSpeed key press detection is weird: sometimes when I press it the move amount increases by 1, sometimes by 2.

Abrackadabra:
I believe this could be caused by the fact that [here](https://github.com/valheimPlus/ValheimPlus/blob/development/ValheimPlus/AdvancedBuildingMode.cs#L212) the translation vectors are being scaled by `Time.deltaTime`. From Unity docs:

The interval in seconds from the last frame to the current one (Read Only).

It appears that this value is not guaranteed at all to be consistent across key presses.

Mixone-FinallyHere:
Solved in #598

PR #598, opened 2021-05-26
Fix ABM translations being inconsistent.

Abrackadabra:
Remove dependency on deltaTime, remove duplicate hook of ABM, pull out a few weeds in the process.


Abrackadabra:
Should fix #455

nxPublic (approved):

PR #600, opened 2021-05-29
Minor refactoring + corrections to enforce naming conventions in C#

fclante:


nxPublic:
1. You modify the projects dependency paths to a fixed path instead of using the $ENV variables.
2. You modify several functions without having tested them, we have written most of the code with clear intend. There is no instance, where we would have copy and pasted vanilla game code instead of calling the function directly. There is reasoning to it, example about that would be line 88 in Inventory.cs. Certain parts of the code are kept more extensive to improve readability in addition.
3. Removing brackets from if statements is not something that makes sense.
4. The naming convention used in our code is called lowerCamelCase.
5. A few logic errors about the usage of our IsEnabled state of the settings configuration, example line 33 in Inventory.cs
6. We are using C# a OOP language, we don't think usage of var is the best standard to resort to.
7. The PR looks untested to us since you would have noticed at least the logic errors i mentioned in 5.
8. The PR isn't even merge ready on submission.
....

I hope you're not offend about us mentioning these issues to you, you obviously spend quite some time on this. 
It's really poor on a lot of ends, we are sorry but we will have to deny this PR. 
If you plan on resubmitting your PR with improved quality, we would appreciate it if you got in touch with us on discord first.

Thank you non the less!

PR #624, opened 2021-08-31
shareablePins not initialized.

IFoundTheLight:
shareablePins in Minimap_AddPin_Patch is causing null exceptions because it isn't initialized to start.


Bigmbrennan:
Please and thank you, console goes insane for me now

prototype464:
Please implement this! Shared pins would be really useful to have working. :>

nxPublic (approved):

Issue #688, opened 2022-03-15
[BUG] Carts are invincible...


prototype464:
- **OS:** Windows 10, 64 bit
- **Valheim Game Version:** Latest
- **V+ Mod Version:** 0.9.9.8

**Describe the bug**
Zero way of destroying placed carts.

**To Reproduce**
1. Set `disableDamageToPlayerStructures` to `true` in `valheim_plus.cfg`
2. Place a cart
3. Try to damage it

**Expected behavior**
Carts should have a separate configuration entry from structures... Why isn't this the case already?

**Troubleshooting**
When I realized carts weren't under the `disableDamageToPlayerBoats` option and were actually under `disableDamageToPlayerStructures`, this upset me because I've been going nuts trying to find out what's causing my carts to be invincible, preventing me from dismantling them and taking them through portals.

The only way to fix this would be to set `disableDamageToPlayerStructures` to `false`, which I shouldn't have to do, because I don't want enemies damaging my bases.

**Pastebin with Valheim Plus config:**
https://pastebin.com/m4uuRzpv

**Other mods installed:**
*BepInEx*
https://valheim.thunderstore.io/package/denikson/BepInExPack_Valheim/
*Valheim Performance Optimizations*
https://www.nexusmods.com/valheim/mods/1360
*Valheim FPS Plus*
https://www.nexusmods.com/valheim/mods/1205
*Custom Raids*
https://www.nexusmods.com/valheim/mods/278
*No Bow Glow*
https://www.nexusmods.com/valheim/mods/1092
*Building Health Display*
https://www.nexusmods.com/valheim/mods/793
*MultiCraft*
https://www.nexusmods.com/valheim/mods/263

increddibelly:
sounds like a feature request to dismantle the cart with the hammer? there's a similar issue around as well #684

nxPublic:
#715

prototype464:
This issue is not fixed, we need to either be able to dismantle boats and carts with hammers or add a 3rd option to toggle whether players can deal damage to them instead of just all entities.

I want to be able to move them across my world via my inventory, but I also don't want random enemies destroying them while I'm not looking.

prototype464:
Bump

Issue #684, opened 2022-03-06
[BUG] Carts and Boats have no way of being dismantled


prototype464:
- **OS:** Windows 10, 64 bit
- **Valheim Game Version:** Latest
- **V+ Mod Version:** 0.9.9.8

**Describe the bug**
After turning on `disableDamageToPlayerBoats` in the config, there isn't a way for the _player_ to damage boats or carts. Valheim does not provide a way to deconstruct these with a hammer, so essentially these are placed permanently.

You can't hit it with a weapon purposely to dismantle it as usual, thus making quick relocation / inventory transport impossible.

**To Reproduce**
1. Set `disableDamageToPlayerBoats` and `disableDamageToPlayerStructures` to `true` in `valheim_plus.cfg`
2. Place a cart, raft, karve, or longship
3. Try to damage it

**Expected behavior**
The player should be able to either dismantle these with a hammer or be able to damage them. The alternate problem here is not being able to disable it specifically for enemies, although the real issue (again) is having no alternative way of destroying them.

**Pastebin with Valheim Plus config:**
https://pastebin.com/m4uuRzpv

**Other mods installed:**
*BepInEx*
https://valheim.thunderstore.io/package/denikson/BepInExPack_Valheim/
*Valheim Performance Optimizations*
https://www.nexusmods.com/valheim/mods/1360
*Valheim FPS Plus*
https://www.nexusmods.com/valheim/mods/1205
*Custom Raids*
https://www.nexusmods.com/valheim/mods/278
*No Bow Glow*
https://www.nexusmods.com/valheim/mods/1092
*Building Health Display*
https://www.nexusmods.com/valheim/mods/793
*MultiCraft*
https://www.nexusmods.com/valheim/mods/263

increddibelly:
I'd say this works as intended 🤔 
sounds like a feature request to be able to dismantle boats with the hammer.

increddibelly:
looking into this and #688 I noticed the [Deconstruct ](https://github.com/valheimPlus/ValheimPlus/blame/799a52a225487cc52e01a3cd77b37aa0e50be048/ValheimPlus/Deconstruct.cs#L12) class was commented out in 9.5 - that's probably intentional. Since it mentions and filters non destructable objects, resurrecting this class might enable a way of tearing down vehicles?

nxPublic:
#715

prototype464:
[...]

Issue #563, opened 2021-05-01
[FEATURE REQUEST] Wagon/Cart Ownership


marvinej:
**Please review the feature requests already submitted to ensure that you are not submitting a duplicate**

**Is your feature request related to a problem? If so, please describe.**
When setting player-built things to invincible it makes it so you cant disassemble and move your cart thru a portal or take it with you easier.

**Describe the solution you'd like**
Setup a Cart-Ownership so only the owner of the cart can destroy it, or if that's not possible add a separate configuration for cart invincibility similar to boats.

nxPublic:
#715

PR #715, opened 2022-11-09
Independent cart damage config

Radvo:
Fixed object handle from "$cart" to "$tool_cart"
Added config support for all damage and water damage for player made carts

Feature requested in :
https://github.com/valheimPlus/ValheimPlus/issues/563
https://github.com/valheimPlus/ValheimPlus/issues/684
https://github.com/valheimPlus/ValheimPlus/issues/688

### DIFF
--- a/README.md
+++ b/README.md
@@ -446,3 +446,4 @@ Please see [CONTRIBUTING.md](/CONTRIBUTING.md) for details on compiling V+ for d
 * An 'Hachidan' N. - https://github.com/ahnguyen09
 * Abra - https://github.com/Abrackadabra
 * Increddibelly - https://github.com/increddibelly
+* Radvo - https://github.com/Radvo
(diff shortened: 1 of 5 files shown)


increddibelly (approved):

increddibelly (approved):

luimen6 (approved):

nxPublic (approved):

prototype464:
My two bug reports were closed however there is still no way to dismantle carts and boats, please either add an option to allow them to be deconstructed with the hammer or another option to allow players to damage them.

stolenvw:
with
`disableDamageToPlayerCarts = false`
i am able to break carts, tested with a sword 

dismantle with hammer did not work

Radvo:
> My two bug reports were closed however there is still no way to dismantle carts and boats, please either add an option to allow them to be deconstructed with the hammer or another option to allow players to damage them.

You can damage both carts and boats if you set the following in the config : 
disableDamageToPlayerBoats = false
disableDamageToPlayerCarts = false


> with `disableDamageToPlayerCarts = false` i am able to break carts, tested with a sword
> 
> dismantle with hammer did not work

That is intended. If you want to make carts immune then set that value to true.

PR #721, opened 2022-11-28
update to .Net Framework v4.8

increddibelly:


increddibelly:
just something vs2022 keeps recommending I do.

increddibelly:
This may not be such a great idea after all. I keep getting MONO errors and using the wrong version of .NET might be the root cause of this. I'll investigate a bit more.

paul-wade:
@increddibelly Another solution that's working for me is to install the 4.5 SDK from the windows 8 SDK https://go.microsoft.com/fwlink/p/?LinkId=226658

tested on windows 11 and vs2022pre

increddibelly:
I believe it is more an issue that it's not _our_ call to decide which framework Valheim uses. Our opinion is irrelevant. Keeping valheimplus' version of .net identical to the version of unity / game code valheim uses makes life easier. I couldnt get it to work, which is also a good reason not to merge :-)

PR #722, opened 2022-11-28
Feature/developer setup tool

increddibelly:
I've been having such a terrible time getting my dev environment set up, I can't even start fixing things. So, I wanted to script the setup. And then I got a bit silly. ~I still want to add the DnSpy debugging tool~ I've added the dnSpy installation and configuration, and a default config. It's starting to become quite useful.

I hope this will bring in more people, since getting started is now a lot easier.


increddibelly:
we should probably mention the original developer for the AssemblyPublicizer. I borrowed some code. to see in case there was anything else we need to publicize, given the recent beta's stream of errors in that area.

SetupDevEnvironment/app.manifest
```diff
@@ -0,0 +1,79 @@
...
+
+        <requestedExecutionLevel  level="asInvoker" uiAccess="false" />
+        <requestedExecutionLevel  level="requireAdministrator" uiAccess="false" />
+        <requestedExecutionLevel  level="highestAvailable" uiAccess="false" />
+
+            Specifying requestedExecutionLevel element will disable file and registry virtualization. 
+            Remove this element if your application requires this virtualization for backwards
+            compatibility.
+        -->
+        <requestedExecutionLevel level="requireAdministrator" uiAccess="false" />
```
increddibelly, line 19: this is required so the app asks for permission to move files around in `C:\Program Files`

paul-wade:
I like the idea, and it's working well for me; up and running quickly. Plus firing up dnspy and the handy config link, excellent work.

increddibelly:
Cool, thanks for the feedback. I didn't get it running with the early test branch builds though, and that's when we need it most.

paul-wade:
@increddibelly Hey why 4.5.1 for the v+ project? Is that to match mono compatibility?

increddibelly:
Initially, vs2022 suggested I upgrade the project to 4.8.2 and I experimented a bit but it failed spectacularly. then I realized that's not my call to make; we depend on valheim and whatever framework they use. Tldr yeah compatibility stuff.

Issue #659, opened 2021-12-27
[FEATURE REQUEST] Increased drops from monsters


ghost:
I might be missing something obvious but there doesn't seem to be an option to increase the amount of items monsters drop. This would be nice to make deers drop more deer hides for example.

PR #772, opened 2023-01-26
Adding LootDrop section with modifiers for loot amount and chance

Grantapher:
This adds a configuration which will change the amount of loot dropped by creatures/monsters when they are slain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,9 @@ We work together closely to make sure all our features are integrated and workin
 * Modify velocity and accuracy of projectiles from monsters.
 * Modify damage and health scaling of monsters in multiplayer based on player count.
 
+### Loot
+* Modify the chance and amount of loot dropped.
+
 # Server
 * Remove password requirement for the server.
 * Modify the maximum amount of players on a server.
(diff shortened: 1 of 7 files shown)


Grantapher:
Related: #659 

I'm not certain `CharacterDrop.DropItems` is the correct place to do this. `Ragdoll` may be better, but I did see there was issue taken in #175. I'm open to suggestions on this.

I couldn't attach dnSpy to debug, so I couldn't dive very deep into how Valheim drops items. I think the `mono-2.0-bdwgc.dll` may be out of date, or I made a mistake (from searching the discord, seems like other people have had issues with this recently as well).

It would be nice to break this down by category of creature/monster dropping the items, but I think in the meantime an overall drop modifier is good to have. Drop *rate* could also possibly be added, but I'm not sure the usefulness of that, nor where to start.

Grantapher:
I ended up exporting the dnSpy assemblies to a visual studio solution and that made it much easier to trace usages of the code. With that, I found a better spot to inject changes to drop amount as well as the drop chance.

I don't see any good way to categorize creatures that aren't overly broad nor requiring a ton of configuration, so I left it as a global loot modifier.

Grantapher:
One thing that I ran into with these changes (not sure 100% if related to loot drop or some other config), is that Oozers also "dropped" a multiplied amount of blobs when killed. I also have all Gathering and Pickables drop rates boosted, but it seems likely due to LootDrop. I'll have to test this some more.

Grantapher:
Added to https://github.com/Grantapher/ValheimPlus/releases/tag/0.9.9.15-alpha6

The oozers dropping extra blobs issue is fixed there.

PR #781, opened 2023-02-05
feat: changed version string patch to include the mod name

### LINKED ISSUE #779: [FEATURE REQUEST] Include ValheimPlus in version string
**Is your feature request related to a problem? If so, please describe.**
ValheimPlus changes the version string to enforce it on all clients, for example `0.212.9` is changed to `0.212.9@0.9.9.11`. It is not always clear for new players which mod is doing this and need time to figure this out.

**Describe the solution you'd like**
The attached string should be changed to include the mod name, so something like `0.212.9-ValheimPlus@0.9.9.11`. Since the version string is changed with each V+ version anyway and there is no parsing involved as far as I'm aware, this should be a non-breaking, minor change that helps new players understanding why they can't connect to a server.

MSchmoecker:
Closes #779.
As already discussed, this makes it more clear for new players and other modders which mod is changing the version string. As this string even changes with every V+ patch version completely and needs to be equal on all players, it is compatible with the current version check.

### DIFF
--- a/ValheimPlus/GameClasses/Version.cs
+++ b/ValheimPlus/GameClasses/Version.cs
@@ -18,7 +18,7 @@ namespace ValheimPlus.GameClasses
                 {
                     if (Configuration.Current.Server.enforceMod)
                     {
-                        __result = __result + "@" + ValheimPlusPlugin.version;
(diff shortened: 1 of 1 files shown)


healiha (approved):
LGTM

Issue #675, opened 2022-02-07
[BUG] Chest emptied when auto deposit honey turned on - intermittent issue


ohrickster:
- **OS:** Windows Client or Server
- **Valheim Game Version:** 0.206.5
- **V+ Mod Version:** [Valheim Plus 0.9.9.7]

Twice I have been playing with the configuration of auto-deposit of honey from hives into a nearby chest.
Most of the time this works flawlessly however, I believe when the chest is full, (chests extended) the chest is emptied the next time it is opened.  The first time this happened was on a windows server (dedicated headless client) and the second time was a solo windows client.

I had just made serpent stew which requires mushrooms and I know I had mushrooms in that chest.   I opened the chest to deposit some more supplies and the only thing left in there was honey.  

**To Reproduce**
I am not sure what is triggering this and I have seen it happen twice but have not been able to manually reproduce it.

Configuration:
https://pastebin.com/dZCPcm3c

**Additional context**

The only other mod I have installed is the "ore support" mod

nxPublic:
Single-player or Multi-player ?

ohrickster:
I've seen this issue happen on a single-player world (local windows client)
and a multi-player dedicated server but have not been able to duplicate it
manually.
I'm attempting to replicate it by filling a chest with stones and leaving a
few slots open for honey deposit to see if it happens again.

bgoosen:
Just saw something very similar to this happen, for the second time now, with a chest next to a smelter with AutoDeposit and AutoFuel on. The chest and smelter hopper are both at their default size.

Both times, it happened when dropping several stacks of iron scrap. Both times I was hardly paying attention and even second guessed what I was seeing, but this time I was sure I had about 3.5 stacks of ore, and suddenly there was only a half stack of ore between the smelter's hopper and the chest. This was on my friend's server, so not sure of it's details.

We're not using any mods other than Valheim Plus.

ghost:
> Just saw something very similar to this happen, for the second time now, with a chest next to a smelter with AutoDeposit and AutoFuel on. The chest and smelter hopper are both at their default size.
> 
> Both times, it happened when dropping several stacks of iron scrap. Both times I was hardly paying attention and even second guessed what I was seeing, but this time I was sure I had about 3.5 stacks of ore, and suddenly there was only a half stack of ore between the smelter's hopper and the chest. This was on my friend's server, so not sure of it's details.
> 
> We're not using any mods other than Valheim Plus.

Same here!

[...]

PR #786, opened 2023-02-19
Refactoring container interactions to avoid overwriting adds/removes

Grantapher:
Updates any container modification to check is the container is in use by a player. If it is, then the container is ineligible for modifications. If it is not, it is set to "in use" while the mod changes the contents, and then sets it back to not in use after. 

These changes aren't very well tested, I did verify that chests being used by players are skipped, but I couldn't reproduce a player being locked out of a chest while the mod changes it (since the time window is so small).

This is mostly an example of what I was thinking in [my comment on #675](https://github.com/valheimPlus/ValheimPlus/issues/675#issuecomment-1431925913) to avoid players putting items into a chest, only for them to be overwritten by the mod also depositing items.

### DIFF
--- a/ValheimPlus/GameClasses/Beehive.cs
+++ b/ValheimPlus/GameClasses/Beehive.cs
@@ -141,10 +141,10 @@ namespace ValheimPlus.GameClasses
                     Inventory cInventory = chest.GetInventory();
                     if (mustHaveItem && !cInventory.HaveItem(item.m_itemData.m_shared.m_name))
                         continue;
-                    if (chest.IsInUse())
+                    if (!chest.IsOwner() || chest.IsInUse())
                         continue;
 
-                    using (InventoryAssistant.lockContainer(chest))
+                    using (InventoryAssistant.LockContainer(chest))
                     {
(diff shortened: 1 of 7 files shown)


MSchmoecker:
`Container.IsInUse()` is not synced in multiplayer, only for the owner of the container this check can ever be true.

The most reliable way is to check for ownership with `Container.IsOwner()` and only access if if this is the case. I also discourage to use `Container.m_nview.ClaimOwnership()`, since this can potentially delete/duplicate items when another player interacts with the chest, even if additional checks are made on the non-owner side.
One of the reliable and not too complex solutions is to ask for ownership, like vanilla does with the `RequestOpen` RPC. There will be a short dely between request and response but the auto deposite doesn't have to be instant anyway.

Grantapher:
Thanks for the info, I'll re-evaluate the code.

Grantapher:
Is the server ever the owner? Or is it always a player who is the owner? Either way, it seems like adding `IsOwner()` checks may be sufficient, but I have a feeling it is more complicated than that.

PR #793, opened 2023-03-16
Fixing compile and patch errors introduced via v0.214.2

Grantapher:
Possibly fixes #789, #790, #791, #792. Unclear if there are any incompatibilities in `Valheim v0.214.2` that didn't cause obvious errors. 

Upgrading `BepInExPack_Valheim` to `5.4.2100` was required. Without it, there were additional errors when using the console or chat, seemingly because of missing `unstripped_corlib` dependencies that were added in the new Valheim version.

I tested the use cases mentioned not working in the linked issues with some exceptions: 
* server sync config
* ABM
* Structural integrity

Given that most of the problems were fixed from changes in mostly unrelated code, I think it is *not terribly risky* to say it was the root cause. 

---

Edit: Since this seems like it may be sitting a while, I uploaded this change to [thunderstore](https://valheim.thunderstore.io/package/Grantapher/ValheimPlus_Grantapher_Temporary/) for folks who install from there and [added a release on my fork](https://github.com/Grantapher/ValheimPlus/releases/tag/0.9.9.12) for folks who install manually. The regular installation instructions will apply to that release.

They both contain [one additional change](https://github.com/Grantapher/ValheimPlus/commit/d01273233c8d29eedb4d037cb785846ded7b77aa) that bumps up the version number, points some version/config checks to my fork, and adds the fact that it is not the main mod to the title screen text. The branch used is the one attached to that commit: [grantapher-development](https://github.com/valheimPlus/ValheimPlus/compare/development...Grantapher:ValheimPlus:grantapher-development).

PR #770, opened 2023-01-23
Add configuration and classes for the Eitr Refinery

shawnwallace:
Uses the same mechanics as the other "smelters", except with different materials.

PR #798, opened 2023-03-22
feat: Add Eitr Refinery

JF10R:
This adds the Eitr Refinery to ValheimPlus, allowing to configure several settings.
This takes into account autoDeposit and autoFuel features.
This pull request requires #793 to work reliably with latest patch.


Grantapher:
Is this duplicating #770?

JF10R:
This is indeed a duplicate in functionality with #770. My bad.

Issue #801, opened 2023-03-25
local defined Keys will not be loaded when connecting to a dedicated server


Grantapher:
I have a bug, that local defined Keys will not be loaded when connecting to a dedicated server. Insted only the default values will be used. I have a fix for it, but can't commit it to your fork (branch: origin\grantapher-development). The button is disabled in VS2022.

I've changed the code in BaseConfig.cs from

```
                if (prop.PropertyType == typeof(KeyCode) && !RPC.VPlusConfigSync.isConnecting)
                {
                    prop.SetValue(this, data.GetKeyCode(keyName, (KeyCode)existingValue), null);
                    continue;
                }
```
to 

```
                if (prop.PropertyType == typeof(KeyCode))
                {
                    prop.SetValue(this, data.GetKeyCode(keyName, (KeyCode)existingValue), null);
                    continue;
                }
```

This is necessary to use FreePlacementRotation without colliding with other keys (e.g. F is used for activateForsaken Powers. This is very frustrating.

---

_Originally posted by @JackTheFragger in https://github.com/valheimPlus/ValheimPlus/issues/793#issuecomment-1483860769_

Grantapher:
Based on commit 149e08e it seems that the code you deleted exists to prevent the server from overriding keys defined in each player's local config. Are you certain you have `Server.serverSyncHotkeys` set to `false`? Are you certain you overwrote the key configuration in your local config?

JackTheFragger:
> 

My server config has the same key definitions. But they will never be read. I've tried key sync enabled and disabled. The properly name 'isConnecting' doedn't say, that it's the KeySync option. I will test it.

JackTheFragger:
I've read the dirty code (you can't call it clean code...). The commit avoided, that the config is ever read. The code is bullshit (sorry).

If HotKeySync in local config is false, the local defined HotKeys will never been read, because there is no other code which reads the config. I can't understand, why the code ever where approved... I will work out a correct method.

JackTheFragger:
The Code is bullshit, because:

The Sync will read the full config and replaces the local config with the server one. If HotKeySync is disabled, the new config will take only the default values, because they will not been overwritten with the local defined... That's why your local config will never run

JackTheFragger:
I've added my tested changes. Now it will peform correctyl. And: now server sync settings will be used for decisions, not local settings.

Grantapher:
[...]

PR #802, opened 2023-03-26
Fix hotkey sync

JackTheFragger:
#801

Local Hotkeys were always been overwritten: by server or by default, because on sync local settings was been ignored
Addition fix: Sync-Options will be used from server config, instead from local config

### DIFF
--- a/ValheimPlus/Configurations/BaseConfig.cs
+++ b/ValheimPlus/Configurations/BaseConfig.cs
@@ -1,15 +1,19 @@
 using IniParser.Model;
 using System.Linq;
+using System.Reflection;
 using UnityEngine;
+using ValheimPlus.RPC;
+using YamlDotNet.Core.Tokens;
+using static CharacterDrop;
 
 namespace ValheimPlus.Configurations
 {
     public interface IConfig
     {
-        void LoadIniData(KeyDataCollection data);
+        void LoadIniData(KeyDataCollection data, string section);
     }
 
-    public abstract class BaseConfig<T> : IConfig where T : IConfig, new()
+    public abstract class BaseConfig<T> : IConfig where T : class, IConfig, new()
     {
 
         public string ServerSerializeSection()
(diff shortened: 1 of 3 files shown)


ValheimPlus/Configurations/BaseConfig.cs
```diff
@@ -25,81 +29,115 @@ public string ServerSerializeSection()
...
         }
 
-        public void LoadIniData(KeyDataCollection data)
+        public void LoadIniData(KeyDataCollection data, string section)
         {
             IsEnabled = true;
+            var thisConfiguration = GetCurrentConfiguration(section);
+            if (thisConfiguration == null)
+            {
+                Debug.Log("Configuration not set.");
```
Grantapher, line 60: Will `Debug.log` appear in user's `player.log`? 

If it does, then we should remove these to not pollute the log.

It it doesn't, then where does it appear so I know when I'm using it 😊

ValheimPlus/Configurations/ConfigurationExtra.cs
```diff
@@ -124,6 +127,15 @@ public static Configuration LoadFromIni(Stream iniStream)
...
                 IniData configdata = parser.ReadData(iniReader);
+                var serverSection = configdata[nameof(Configuration.Server)];
+                var serverSyncsConfig = serverSection.GetBool(nameof(ServerConfiguration.serverSyncsConfig));
+                Debug.Log($"ServerSyncsConfig = {serverSyncsConfig}");
+
+                if (!serverSyncsConfig) return Configuration.Current;
+
+                var serverSyncsHotkeys = serverSection.GetBool(nameof(ServerConfiguration.serverSyncHotkeys));
+                Debug.Log($"ServerSyncsHotkeys = {serverSyncsConfig}");
+                ReadHotKeys = serverSyncsHotkeys;
```
Grantapher, line 138: As mentioned in #801, this should be a client side setting.

Grantapher (changes requested):
Thanks for making this PR, it definitely is odd to me that players could ignore the `serverSyncsConfig` settings.

